PR #6, opened 2022-07-25
Add command to manage Extensions/skins

amalpaul54111:


jeffw16:
To-do:
- Only list directories rather than files too (so we can't enable the "README" "extension" and "skin")
- Add a comment line saying "This file was generated by Canasta"; the CLI only does stuff if that exists in the file
- Add multiple extensions and skins at the same time, with names separated by commas
- Ensure idempotency so that existing files shouldn't be overridden

Wish list:
- Add adjacency list of dependencies
- Support for adding/removing user-added extensions? (Let's not worry about this for now)

internal/extensionsskins/extensionsskins.go
```diff
@@ -0,0 +1,68 @@
...
+	return false
+}
+
+func List(instance logging.Installation, constants Item) {
+	fmt.Printf("Available %s:\n", constants.Name)
+	fmt.Print(orchestrators.Exec(instance.Path, instance.Orchestrator, "web", "cd $MW_HOME/"+constants.RelativeInstallationPath+" && find * -maxdepth 0 -type d"))
+}
+
+func CheckInstalled(name string, instance logging.Installation, constants Item) (string, error) {
+	output := orchestrators.Exec(instance.Path, instance.Orchestrator, "web", "cd $MW_HOME/"+constants.RelativeInstallationPath+" && find * -maxdepth 0 -type d")
```
jeffw16, line 32: please add spaces between `+` for all instances in which string concatenation occurs in this PR

jeffw16 (changes requested):

internal/extensionsskins/extensionsskins.go
amalpaul54111: As Yaron mentioned in the other pull request, it's something done by the `gofmt` the code formatting tool for golang.

yaronkoren:
This looks fine to me - the only issue I see is that "canasta-extension" and "canasta-skin" should instead be "Canasta extension" and "Canasta skin". Normally I wouldn't care about small typos like this, but it shows up a bunch of times.

jeffw16:
Agreed with Yaron ^

jeffw16 (approved):

PR #7, opened 2022-07-25
Add Restart Command

amalpaul54111:


cmd/restart/restart.go
```diff
@@ -0,0 +1,55 @@
...
+			logging.SetVerbose(verbose)
+			if instance.Id == "" && len(args) > 0 {
+				instance.Id = args[0]
+			}
+			fmt.Println("Restarting Canasta")
+			err := Restart(instance)
+			if err != nil {
+				log.Fatal(err)
+			}
+			fmt.Println("Started")
```
jeffw16, line 30: Shouldn't this not appear if `err != nil`?

amalpaul54111: Currently, the restart command would only be printing out errors concerned with the CLI, such as if the canasta ID is not found. If there are any errors that caused the docker-compose to return an exit code other than 0, the CLI would print the output of the command that caused it to exit, and exit the CLI itself.

In such a case I am not sure, if it returned an error, whether to display that the instance has been started.

yaronkoren: This looks good to me - what's the issue with this line of code?

jeffw16 (approved):

PR #13, opened 2022-09-03
Add checks to make sure canasta's requirements are met

### LINKED ISSUE #12: Canasta should verify dependencies before doing anything
The cli should verify that docker-compose, etc are available when it is invoked with any command that will use them.

For instance: `canasta help` does not require any external tools and it should (continue to) just work.

However `canasta create` does require the installation of [docker-compose](https://docs.docker.com/compose/) and it should refuse to run if it cannot find it.  Otherwise, you end up with this sort of frustrating dialog:
```
$ sudo ./canasta create
Enter Wiki Name: Test
Enter Canasta ID: test
Enter admin name: WikiSysop
Enter the admin password (Press Enter to autogenerate the password): 
Setting up Canasta
exec: "docker-compose": executable file not found in $PATH
$

hexmode:
* Check that we're running as root (though, I'm not sure it is necessary when you are running a
  non-root container system)
* Check that we have docker-compose installed before attempting create
* Oh, and btw, update gitignore for emacs files

fixes #12


jeffw16 (approved):
LGTM

jeffw16:
@amalpaul54111 Any comments for this PR?

amalpaul54111:
Since docker-compose is not only used by the create command, I think it might be a better idea to add that `docker-compose` check to maybe within the `internal/orchestrators/orchestrators.go` inside a `init()`. I guess that would make more sense. So on any import of the that file, dependencies could be checked.

hexmode:
> Since docker-compose is not only used by the create command, I think it might be a better idea to add that `docker-compose` check to maybe within the `internal/orchestrators/orchestrators.go` inside a `init()`.

Agreed.  I also have other ideas on this that I'll add.

amalpaul54111:
That's great!. Thank You.

jeffw16:
Ping me when another code review is needed!

hexmode:
@jeffw16 ping?

jeffw16 (approved):
LGTM

yaronkoren:
@hexmode - could you resolve the conflicts in these three files? They seem pretty straightforward. (Obviously it would have been better to just merge this in a month ago, but... well, the past is the past.)

hexmode:
@yaronkoren updated

jeffw16:
LGTM, should we merge?

yaronkoren:
I would say yes! I don't know if I feel comfortable merging this in myself, since I haven't tested it, but... it seems good.

PR #19, opened 2022-09-05
Delete installation on failure

### LINKED ISSUE #14: When canasta has failed, but created directories for an install, it can't run again against the same `create` without error
Example:
```
$ ls -l test
total 24
drwxr-xr-x. 3 root root   86 Aug 19 01:15 config
-rw-r--r--. 1 root root  545 Aug 19 01:15 docker-compose.override.yml
...
$ sudo ./Canasta-CLI-Go create
Enter Wiki Name: test
Enter Canasta ID: test
Enter admin name: test
Enter the admin password (Press Enter to autogenerate the password): 
Setting up Canasta
fatal: destination path '/home/mah/Canasta-CLI/test' already exists and is not an empty directory.
$

### PR DESCRIPTION
**Pull request by amalpaul54111**
Fixes #14
So when a create/import command fails, `canasta` CLI asks the user whether to keep the config files or not.

### DIFF
--- a/cmd/create/create.go
+++ b/cmd/create/create.go
@@ -33,18 +33,19 @@ func NewCmdCreate() *cobra.Command {
 			fmt.Println("Setting up Canasta")
 			if err = createCanasta(canastaInfo, pwd, path, orchestrator); err != nil {
 				fmt.Print(err.Error(), "\n")
-				fmt.Println("A fatal error occured during the installation\nDo you want to delete the files related to it? (y/n)")
+				fmt.Println("A fatal error occured during the installation\nDo you want to keep the files related to it? (y/n)")
 				scanner := bufio.NewScanner(os.Stdin)
 				scanner.Scan()
 				input := scanner.Text()
-				if input == "y" {
-					installationDir := path + "/" + canastaInfo.Id
-					fmt.Println("Removing containers")
-					orchestrators.DeleteContainers(installationDir, orchestrator)
-					fmt.Println("Deleting config files")
-					orchestrators.DeleteConfig(installationDir)
+				if input == "y" || input == "Y" || input == "yes" {
 					logging.Fatal(fmt.Errorf("Exiting"))
 				}
+				installationDir := path + "/" + canastaInfo.Id
+				fmt.Println("Removing containers")
+				orchestrators.DeleteContainers(installationDir, orchestrator)
+				fmt.Println("Deleting config files")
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

**Comment by hexmode** [comment]
Tested.

Looks pretty good to me.  I would prefer a CLI option to keep and then change the default to delete, but if @jeffw16 or @yaronkoren like this implementation, it works.

**Comment by yaronkoren** [comment]
@hexmode - great!

I think this is good - even if the plan is to add a "keep on failure" option, this is a step in the right direction anyway. (A big step.)

**Comment by jeffw16** [comment]
I agree with @hexmode: it should be delete by default.

File: cmd/create/create.go [hunk 1]
```diff
@@ -30,8 +32,19 @@ func NewCmdCreate() *cobra.Command {
...
 			fmt.Println("Setting up Canasta")
 			if err = createCanasta(canastaInfo, pwd, path, orchestrator); err != nil {
-				orchestrators.Delete(path, orchestrator)
-				logging.Fatal(err)
+				fmt.Print(err.Error(), "\n")
+				fmt.Println("A fatal error occured during the installation\nDo you want to delete the files related to it? (y/n)")
+				scanner := bufio.NewScanner(os.Stdin)
+				scanner.Scan()
+				input := scanner.Text()
+				if input == "y" {
```
jeffw16: Please change to delete by default, keep only by explicit request

jeffw16 (changes requested):

cmd/create/create.go
jeffw16: Also I'm pretty sure we can do better than simply checking for a lowercase `y` or `n`. There's more possible cases, such as `yes` or `no` or `Y` or `N`.

amalpaul54111: > Also I'm pretty sure we can do better than simply checking for a lowercase `y` or `n`. There's more possible cases, such as `yes` or `no` or `Y` or `N`.

Yeah, sure. 

> Please change to delete by default, keep only by explicit request

Okay, also should there be a CLI option to keep the files, so it does not prompt?

yaronkoren: I would say, in the interests of time (and keeping this patch manageable), don't worry about adding an option to keep the files - it would be nice to have, but it can always be added later.

jeffw16: I agree with @yaronkoren

jeffw16 (approved):

PR #29, opened 2022-10-18
Only call "show-progress" in wget during install if that param exists

### LINKED ISSUE #27: Remove show progress flag from wget for CLI (or check wget version)
Older versions of wget don't have support for the progress bar. We need to remove this option or only call it if it's above a certain version of wget.

yaronkoren:
Closes #27.


jeffw16 (approved):
LGTM

PR #33, opened 2022-10-26
Add build specification for Linux amd64 for build script

jeffw16:


hexmode:
I don't like this change.  Previously, I used this script to build on a system whose architecture is `aarch64`.  With this change, the build still produces an executable, but it isn't usable on the `aarch64` architecture.

`file` output with this change:
```
$ file Canasta-CLI-Go 
Canasta-CLI-Go: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=kd99VGlX0VmBfDCHDqFU/bL9H40Kdr_Kwu9tELQF0/Bw_tZF7hmNc__FbnWhGS/jmkKILVS_SUyTcmVbw-b, not stripped
```

`file` output without this change:
```
Canasta-CLI-Go: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, Go BuildID=Juo1u-3gJJ00YmoD2EQZ/B3WA1B8YDYzSnYw8YFzz/cHP6t1HSFqs5nyI70iIg/HhuVoEM19jYNUa3IpRN3, not stripped
```

This change would make the CLI produced with `build.sh` un-usable on OS X or FreeBSD or..., as well.

jeffw16:
Great point, will respond later

jeffw16:
1. This change was made when we didn't know if AArch64 support would be feasible for Canasta
2. My computer has since shifted from an Intel Mac to an ARM Mac (unfortunately)
3. I of course support changing this to build on multi-platform, but I suppose the best way to do so now is to let GitHub Actions create the binaries for us with a CI/CD workflow rather than having each person download their own and compile it. Unfortunately this doesn't work too well I don't think because GitHub doesn't have ARM runners yet.

hexmode:
Right now, the necessity of this is confusing to me.  If you leave GOARCH un-specified and the user has go installed, this script will build executables that work with the user's platform.

If you are trying to build for different targets, then that can be done by changing the envvars when you call the the script. Worrying about the runner shouldn't be an issue since Golang [comes with support for all architectures built in](http://dave.cheney.net/2015/03/03/cross-compilation-just-got-a-whole-lot-better-in-go-1-5). For instance, here is the current main branch on my Linux laptop:
``` console
$ uname -m
x86_64
$ file ./Canasta-CLI-Go 
./Canasta-CLI-Go: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, Go BuildID=ilpUl6gCUzcDGdQUt0tT/_ZenVQF_oe60ZRV0bYk3/2yYCwCBEsxKJTQYlPaQ5/6RwdrXhJLjOJbvfYfJLf, with debug_info, not stripped
$ GOARCH=arm ./build.sh
$ file ./Canasta-CLI-Go 
./Canasta-CLI-Go: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), statically linked, Go BuildID=c90t4YF_C8PzR2oZQKoo/uTXXFT9zmOYyvXqU7dr9/T0kVksfMBdo1eeToYChq/aAEZnXAe9NfEbDlJAdOy, with debug_info, not stripped
```
Since [github actions allow you to set environment variables](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs), I think this would work. If that isn't sufficient, then creating a script that creates multiple binaries by running `build.sh` inside a loop that sets GOARCH to the targets desired seems like a good fallback.

jeffw16:
Got it, thanks @hexmode, let's get rid of this PR and do it your way

yaronkoren:
Per year-old request, I'm closing this!

PR #35, opened 2022-10-30
feat: Add option to choose version to install

### LINKED ISSUE #34: Add option to choose version to install
Users should be able to choose what version they want to install, as well as any derivative images.

For instance, I might not want Canasta 1.2.0 if it is the most recently released version. Instead, I could want Canasta 1.1.2 for backwards compatibility reasons. Or, I might want to use the hardened version of Canasta or even Open CSP on Canasta. None of this is possible at the moment.

ha36d:
fixes CanastaWiki/Canasta-CLI#34

Signed-off-by: Hamed Faramarzi <hamed.faramarzi@gmail.com>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,10 +1,67 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Canasta CLI installer
 # Requirements Docker Engine 18.06.0+ and DockerCompose 
 
-echo "Downloading Canasta CLI latest release"
-canastaURL="https://github.com/CanastaWiki/Canasta-CLI/releases/latest/download/canasta"
+usage() {
+  cat << EOF >&2
+Usage: $PROGNAME [-i] [-v <ver>]
+
+-v <ver>: install specific version
+      -i: interactive version
+       *: usage
+EOF
+  exit 1
+}
+
+while getopts v:i opt; do
+  case ${opt} in
+  i)
+    echo "interactive mode"
+    INTERACTIVE=1
+    ;;
+  v)
+    VERSION=$OPTARG
+    ;;
+  *)
+    usage
+    ;;
+  esac
+done
+
+if [[ -n ${INTERACTIVE-} ]]; then
+  echo "-----"
+  echo "Checking the Version..."
+  echo "Which version do you want to install?"
+  echo "   1) Default: latest"
+  echo "   2) Custom"
+  until [[ $VERSION_CHOICE =~ ^[1-2]$ ]]; do
+  	read -rp "Version choice [1-2]: " -e -i 1 VERSION_CHOICE
+  done
+  case $VERSION_CHOICE in
+  1)
+    VERSION="latest"
+  	;;
+  2)
+    until [[ $VERSION =~ ^(([0-9]{1,3}[\.]){2}[0-9]{1,3}).* ]]; do
+      read -rp "Version (e.g. 1.2.0): " -e VERSION
+  	done
+  	;;
+  esac
+fi
+
+if [[ -z ${VERSION-} ]]; then
+  VERSION="latest"
+  echo "No version have been specified, latest will be used."
+fi
+
+if [[ ${VERSION} == "latest" ]]; then
+  canastaURL="https://github.com/CanastaWiki/Canasta-CLI/releases/latest/download/canasta"
+else
+  canastaURL="https://github.com/CanastaWiki/Canasta-CLI/releases/download/v${VERSION}/canasta"
+fi
+
+echo "Downloading Canasta CLI ${VERSION} release"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -1,17 +1,45 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
jeffw16, line 1: Any specific reason to change the interpreter path?

install.sh
```diff
@@ -1,17 +1,45 @@
...
+	;;
+esac
+
+echo "Downloading Canasta CLI ${VERSION} release"
 # The show-progress param was added to wget in version 1.16 (October 2014).
 wgetOptions=$(wget --help)
 if [[ $wgetOptions == *"show-progress"* ]]
 then
   wget -q --show-progress $canastaURL
+  if [ $? -ne 0 ]; then
```
jeffw16: Can we avoid duplicating this if statement

install.sh
```diff
@@ -1,17 +1,45 @@
...
+	read -rp "Version choice [1-2]: " -e -i 1 VERSION_CHOICE
+done
+case $VERSION_CHOICE in
+1)
+  VERSION="latest"
+	canastaURL="https://github.com/CanastaWiki/Canasta-CLI/releases/latest/download/canasta"
+	;;
+2)
+  until [[ $VERSION =~ ^(([0-9]{1,3}[\.]){2}[0-9]{1,3}).* ]]; do
+    read -rp "Version (i.e. 1.2.0): " -e VERSION
```
jeffw16: Change `i.e.` to `e.g.`

install.sh
```diff
@@ -1,17 +1,45 @@
...
+esac
+
+echo "Downloading Canasta CLI ${VERSION} release"
 # The show-progress param was added to wget in version 1.16 (October 2014).
 wgetOptions=$(wget --help)
 if [[ $wgetOptions == *"show-progress"* ]]
 then
   wget -q --show-progress $canastaURL
+  if [ $? -ne 0 ]; then
+    echo "Canasta CLI with the version you have chosen, does not exist."
```
jeffw16: Please change to `The version you have specified is not a valid version of the Canasta CLI.`

jeffw16 (changes requested):
Thanks for the contribution!

install.sh
ha36d: yes, because the `-i` option in `read` is not supported in Mac Bash. It should be updated, and everyone update it via brew. It will not impact linux ones. `/usr/bin/env` has the benefit of looking for whatever the default version of the program is in the current environment.

install.sh
ha36d: sure

install.sh
ha36d: Done.

install.sh
ha36d: Done

install.sh
```diff
@@ -1,10 +1,30 @@
...
+#!/usr/bin/env bash
 
 # Canasta CLI installer
 # Requirements Docker Engine 18.06.0+ and DockerCompose 
 
-echo "Downloading Canasta CLI latest release"
-canastaURL="https://github.com/CanastaWiki/Canasta-CLI/releases/latest/download/canasta"
+echo "-----"
+echo "Checking the Version..."
+echo "Which version do you want to install?"
```
jeffw16: It's important to keep a non-interactive version of this script. The behavior should be as follows:

- By default, if this script is called normally, the latest version is installed.
- If a flag `--version` or `-v` is passed in with the desired version, install the specified version.
- If a flag `--interactive` or `-i` is passed in, prompt what exists right now.

jeffw16 (changes requested):

install.sh
ha36d: Done.

jeffw16 (approved):

yaronkoren:
Note: this is a fix for issue #34.

yaronkoren:
@ha36d - I'm sorry for the very long delay in reviewing this patch; I wasn't paying enough attention to CLI pull requests before. This overall looks good - but I think the "interactive mode" is unnecessarily complex, and doesn't make the "version" feature any easier to use, I don't think. I think just the "-v" flag should be enough; people will understand what it means. Could you simplify this patch? I hope you are still willing to do some work on this...

yaronkoren:
Actually, another comment: calling the flag "-v" might not be all that clear, since people might think it refers to the version of the Canasta CLI. (Or even the version of the installer, for that matter...) Could you change it to something like "-canasta_version"?

yaronkoren:
Closing/rejecting this PR - it still has some issues, and in the meantime it has been superseded by #77.

PR #36, opened 2022-11-02
Add Upgrade command

amalpaul54111:
Add the “canasta upgrade” command, which will update the canasta installation to the latest version.

Currently, it runs “git pull origin main” inside the configuration folder, which is a git clone of the original stack repo. So by pulling the updates and restarting the containers, it gets updated.

### DIFF
--- a/cmd/upgrade/upgrade.go
+++ b/cmd/upgrade/upgrade.go
@@ -1,0 +1,66 @@
+package upgrade
+
+import (
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/CanastaWiki/Canasta-CLI-Go/internal/canasta"
+	"github.com/CanastaWiki/Canasta-CLI-Go/internal/git"
+	"github.com/CanastaWiki/Canasta-CLI-Go/internal/logging"
+	"github.com/CanastaWiki/Canasta-CLI-Go/internal/orchestrators"
+)
+
+var (
+	pwd      string
+	err      error
+	instance logging.Installation
+)
+
+func NewCmdCreate() *cobra.Command {
+	var upgradeCmd = &cobra.Command{
+		Use:   "upgrade",
+		Short: "Upgrade a Canasta installation to the latest version",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if instance.Id == "" && len(args) > 0 {
+				instance.Id = args[0]
+			}
+			if err := Upgrade(instance); err != nil {
+				return err
+			}
+			return nil
+		},
+	}
+	if pwd, err = os.Getwd(); err != nil {
+		log.Fatal(err)
+	}
+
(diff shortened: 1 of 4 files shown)


bawolff (changes requested):
So I assume that upgrading canasta versions can sometimes upgrade MediaWiki?

If so, I would recommend running the update.php maintenance script as part of this. I would also recommend `touch` -ing LocalSettings.php, as that triggers a cache flush in MediaWiki (for some types of caches).

Other than that, this looks good.

amalpaul54111:
Yeah sure, that can be done.

amalpaul54111:
Should we explicitly run the “update.php” maintenance script since we are already doing a **restart** and as the Canasta container image does run the “update.php” scrip as part of the **run_apache.sh** script on container startup. 

Also would a restart of the container flush cache or is it necessary to do  “touch LocalSettings.php”?

bawolff:
> Should we explicitly run the “update.php” maintenance script since we are already doing a restart and as the Canasta container image does run the “update.php” scrip as part of the run_apache.sh script on container startup.

Oh, i didn't realize that. Given that, not running update.php here makes sense.

> Also would a restart of the container flush cache or is it necessary to do “touch LocalSettings.php”?

Touching LocalSettings.php will trigger browser cache to flush (i.e. prevent sending 304 Not Modified for stuff pre-touch), flush parser cache, and flush file cache. Restarting the container will probably flush parser cache (Assuming we are not using CACHE_DB) but I don't believe it will flush browser cache (And i don't think file cache is relevant to canasta). So its probably still a good idea, but  also not critical.

amalpaul54111:
Okay, could add 'touch LocalSettings.php', just wanted to clarify.

cmd/upgrade/upgrade.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	//Restarting the containers
+	if err = orchestrators.StopAndStart(instance.Path, instance.Orchestrator); err != nil {
+		return err
+	}
+
+	//Touch LocalSettings.php
+	fmt.Print("Running 'touch LocalSettings.php' to flush cache\n")
+	execute.Run(instance.Path, "touch", "config/LocalSettings.php")
```
bawolff, line 68: This won't work because that's not the real LocalSettings.php file, so MediaWiki won't detect the change. Would probably have to touch `/var/www/mediawiki/w/LocalSettings.php` inside the web container for this to work.

Although this does raise an interesting point, that when ordinary users edit LocalSettings.php in Castana, they are going to edit config/LocalSettings.php, which won't have the cache clearing affect that would happen in a normal mediawiki install.

bawolff (changes requested):

yaronkoren:
This code was more-or-less merged in, with some modifications, in #102. @amalpaul54111 - thank you for your work on this!

PR #44, opened 2022-11-08
Save password to .admin-password

### LINKED ISSUE #40: Automatically-generated passwords don't show up anywhere?
If you get an automatically-generated password for the admin account by leaving out the "-p" flag in "canasta create", the password that's generated doesn't seem to appear anywhere - so as far as I know it's lost forever. (If it does get stored somewhere, it would be good to know where - so that this can be noted in the documentation, and probably the CLI output as well.)

amalpaul54111:
Fixes #40 
Save the automatically generated password to ```.admin.password``` inside the configuration folder.

### DIFF
--- a/internal/mediawiki/mediawiki.go
+++ b/internal/mediawiki/mediawiki.go
@@ -50,6 +50,17 @@ func Install(path, orchestrator string, canastaInfo canasta.CanastaVariables) (c
 		if err != nil {
 			return canastaInfo, err
 		}
+		// Save automatically generated password to .admin.password inside the configuration folder
+		fmt.Printf("Saving password to %s/.admin.password\n", path)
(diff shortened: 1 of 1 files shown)


yaronkoren:
This looks good! I'm looking forward to hearing what others have to say. Do you think ".admin.password" is a better name than ".admin-password"?

amalpaul54111:
Oops!, I did misread  “.admin-password” as “.admin.password”.
".admin-password” would be better for sure.

jeffw16 (approved):
LGTM

freephile:
Is this documented? Where should the feature be documented?

If you choose to have Canasta generate an admin password for you, you will find it in a file named .admin-password

PR #47, opened 2022-12-05
Use bash shell when executing commands internally

### LINKED ISSUE #32: Explicitly set shell used to bash
In Ubuntu, the following error occurs upon running `sudo canasta create`:

```
$ sudo canasta create -i test -n smwcon.mywikis.eu -w "SMWCon Canasta" -a admin -o docker-compose
/usr/local/bin/canasta: 1: Syntax error: "(" unexpected
```

This is because of:
- https://unix.stackexchange.com/questions/45781/shell-script-fails-syntax-error-unexpected
- https://wiki.ubuntu.com/DashAsBinSh

The mitigation for this bug is to change Ubuntu's default shell from dash back to bash, but the real resolution for this problem is to ensure Go will call commands with bash.

amalpaul54111:
Fix #32


jeffw16 (approved):
LGTM

PR #48, opened 2022-12-09
Fix config dir

### LINKED ISSUE #42: Create .config directory if it's needed and doesn't exist
The function init() in config.go calls the function getConfigDir(), which in turn looks for the directory `.config/canasta` - and this call seems to always fail (and display an error message), because the directory doesn't exist.

It appears that all of this code was added in this patch, which was committed about two weeks ago:

https://github.com/CanastaWiki/Canasta-CLI/pull/13/files

@hexmode - can you explain what is going on with this code? Should the directory be getting created at some point? Or should this part of the code just be removed?

amalpaul54111:
Fix #42
1. Remove error saying missing "/root/.config" directory which is not used.
2. Create ".config" in users home directory if it does not exist.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -11,6 +11,7 @@ import (
 	"path"
 	"syscall"
 
+	"github.com/CanastaWiki/Canasta-CLI-Go/internal/logging"
 	"github.com/kirsle/configdir"
 )
 
@@ -31,7 +32,7 @@ var (
 func Exists(canastaId string) bool {
 	err := read(&existingInstallations)
 	if err != nil {
-		log.Fatal(err)
+		logging.Fatal(err)
 	}
 	return existingInstallations.Installations[canastaId].Id != ""
 }
@@ -39,7 +40,7 @@ func Exists(canastaId string) bool {
 func ListAll() {
 	err := read(&existingInstallations)
 	if err != nil {
-		log.Fatal(err)
+		logging.Fatal(err)
 	}
 	fmt.Printf("Canasta ID\tInstallation Path\t\t\t\t\tOrchestrator\n\n")
 	for _, installation := range existingInstallations.Installations {
@@ -78,10 +79,10 @@ func Delete(canastaID string) error {
 	if Exists(canastaID) {
 		delete(existingInstallations.Installations, canastaID)
 	} else {
-		log.Fatal(fmt.Errorf("Canasta installation with the ID doesn't exist"))
+		logging.Fatal(fmt.Errorf("Canasta installation with the ID doesn't exist"))
 	}
 	if err := write(existingInstallations); err != nil {
-		log.Fatal(err)
+		logging.Fatal(err)
 	}
 
 	return nil
@@ -90,7 +91,7 @@ func Delete(canastaID string) error {
 func write(details Canasta) error {
 	file, err := json.MarshalIndent(details, "", "	")
 	if err != nil {
-		log.Fatal(err)
+		logging.Fatal(err)
 	}
 	return ioutil.WriteFile(confFile, file, 0644)
 }
@@ -98,7 +99,7 @@ func write(details Canasta) error {
 func read(details *Canasta) error {
 	data, err := os.ReadFile(confFile)
 	if err != nil {
-		log.Fatal(err)
+		logging.Fatal(err)
 	}
 	err = json.Unmarshal(data, details)
 	return err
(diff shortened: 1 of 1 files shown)


yaronkoren:
@hexmode - what do you think, does this look like correct logic?

internal/config/config.go
```diff
@@ -107,28 +107,32 @@ func read(details *Canasta) error {
...
 		errReport := fmt.Errorf("Unable to get the current user: %s", err)
 		log.Fatal(errReport)
 	}
+
 	if currentUser.Username == "root" {
 		dir = directory
-	} else if currentUser.Username != "root" && exists != true {
-		msg := fmt.Sprintf("Using %s for configuration...", dir)
-		log.Print(msg)
+	} else if currentUser.Username != "root" {
```
hexmode: this should just be else

internal/config/config.go
```diff
@@ -119,7 +119,14 @@ func GetConfigDir() string {
 		dir = directory
 	} else if currentUser.Username != "root" {
 		fi, err := os.Stat(dir)
-		if err != nil {
+		if os.IsNotExist(err) {
+			var configDir string = ".config"
```
hexmode: This should use the current value of dir.

internal/config/config.go
```diff
@@ -119,7 +119,14 @@ func GetConfigDir() string {
...
+			log.Print(fmt.Sprintf("Creating %s\n", configDir))
+			err = os.Mkdir(configDir, os.ModePerm)
```
hexmode: This should use [os.MkdirAll](https://pkg.go.dev/os#MkdirAll).

internal/config/config.go
```diff
@@ -119,7 +119,14 @@ func GetConfigDir() string {
...
+			if err != nil {
+				log.Fatal(err)
+			}
```
hexmode: exists should be set to true here

hexmode (changes requested):

internal/config/config.go
amalpaul54111: Oh, sorry, that's right.

amalpaul54111:
@hexmode I have made the requested changes. 
Since the `config.init()` is being executed even before setting the verbose, used the `log.Print()` inside both the `config.GetConfigDir()` and `config.init()`
Also removed the repeated section which verified the configuration directory existed.

internal/config/config.go
```diff
@@ -107,8 +108,27 @@ func read(details *Canasta) error {
...
+	}
+
+	if currentUser.Username == "root" {
+		dir = directory
+	}
+
+	fi, err := os.Stat(dir)
+	if os.IsNotExist(err) {
+		log.Print(fmt.Sprintf("Creating %s\n", dir))
+		err = os.MkdirAll(dir, os.ModePerm)
```
bawolff, line 126: os.ModePerm (= o777) seems like the incorrect permissions here.

Assuming this config file is meant to be per-user with no secret data, I would expect permissions to be o755

bawolff (changes requested):
I'm not sure I understand the intention of this code.

If the goal is that every user has a separate config file, then i think it makes more sense for the root user to use /root/.config/canasta . Normally I would expect a file in /etc to be shared by all users. That said, if this was to be shared by all users, I would probably expect it in /var/local since this is state related data and not configuration related data.

Anyways, I would suggest just simplifying this to not have the special logic for the root user.

yaronkoren:
@bawolff - by "this code", do you mean this patch, or the underlying .config file feature? If it's this latter, that discussion probably makes more sense in a separate issue/PR - though we should definitely have that discussion.

bawolff:
> @bawolff - by "this code", do you mean this patch, or the underlying .config file feature? If it's this latter, that discussion probably makes more sense in a separate issue/PR - though we should definitely have that discussion.

I meant the underlying, put config in /etc if you are root feature (Or perhaps the idea of having different config files for different users in general).  And yes, that sounds fair.

My only other objection was that the directory should probably be 0755 or 0700 depending on your level of paranoia not 0777,  however that also seems to be copied over from the previous version of the file, and is probably a separate issue from the thing this patch is trying to fix. So with that in mind, this patch looks good, and I'll file a separate issue to further discuss the config file thing.

bawolff (approved):

yaronkoren (approved):

tosfos (approved):

internal/config/config.go
tosfos: To be addressed in #64

PR #49, opened 2022-12-11
Add Maintenance script command

### LINKED ISSUE #43: Need CLI command to call any MW "maintenance" script
It would be great to have a command that simply calls any script in the core MW maintenance/ directory - and lets you pass any parameters directly to that script - so that you don't have to go into the Docker container just to call, say, runJobs.php.

I know there's already a command called "canasta maintenance", so this would need a different name than that - what about "canasta maintenance script"? If that were done, perhaps the command now at "canasta maintenance" would be better made a subcommand too, like "canasta maintenance update" - which, looking at the code, is already kind of the name of this command anyway...

amalpaul54111:
1. Move existing `maintenance` command to `maintenance update`.
2. Add `maintenance script` command to run any PHP scripts inside core MW maintenance/ directory. 

Fixes #43.

### DIFF
--- a/cmd/maintenanceUpdate/maintenance.go
+++ b/cmd/maintenanceUpdate/maintenance.go
@@ -21,6 +21,7 @@ func NewCmdCreate() *cobra.Command {
 	}
 
 	maintenanceCmd.AddCommand(updateCmdCreate())
+	maintenanceCmd.AddCommand(scriptCmdCreate())
 	if pwd, err = os.Getwd(); err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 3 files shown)


bawolff (approved):
I think this is good, but some things to note:

So ideally this wouldn't require you to quote the full command, and instead would recognize that all arguments after the script name are supposed to be commands. However i don't think that's a blocker to merging.

This allows command injection. I assume that's not considered a concern in context (e.g. You can do `canasta maintenance script "showJobs.php;id"` . However given the command is to run arbitrary scripts anyways, it doesn't feel like that matters much).

yaronkoren (approved):

Issue #26, opened 2022-10-13
Add Git as prereq to code and documentation


jeffw16:
The CLI uses Git to manage the stack repo. We need to add checks for it in the CLI plus documentation.

yaronkoren:
I would say this is fixed.

PR #52, opened 2022-12-15
Test for git

gitmapd:
Added test for git and replaced "which", there are distributions that "which" doesn't exist, it's not portable.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -30,7 +30,7 @@ else
      echo "Git was found on the system"
 fi
 
-loc=$(command -v docker)
+loc=$(which docker)
 if [ -z $loc ]
 then
     echo "Docker is not installed; please follow the guide at https://docs.docker.com/engine/install/ to install it."
@@ -41,7 +41,7 @@ else
     echo "Docker appears to be installed at $loc but is not executable; please check permissions."
 fi
 
-loc=$(command -v docker-compose)
+loc=$(which docker-compose)
 if [ -z $loc ]
 then
     echo "Docker Compose is not installed; please follow the guide at https://docs.docker.com/compose/install/ to install it."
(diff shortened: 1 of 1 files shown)


gitmapd:
I forgot to change the shebang to #!/usr/bin/env bash .

install.sh
```diff
@@ -18,6 +18,18 @@ echo "Installing Canasta CLI"
...
+GIT_IS_AVAILABLE=$?
+
+if [ $GIT_IS_AVAILABLE -ne 0 ]; 
+then echo "Git was not found, please install before continuing.";
+     exit; 
+else
+     echo "Git was found on the system"
+fi
+
 loc=$(which docker)
```
jeffw16, line 33: `which` and the lines below it that use the `$loc` variable don't seem to be removed from here.

jeffw16 (dismissed):

gitmapd:
Hi, I apologize but i don't get your sentence. Thanks.

install.sh
jeffw16: You said you replaced `which` because it's not portable, but your new code still contains `which`.

gitmapd:
I know i i think you got to see the comment before I removed it, but Yaron
told me to make a new patch for it. There is another patch to change the
shebang, do you want me to make a patch with which + the shebang changed?

yaronkoren:
Note: this fixes issue #26.

bawolff (approved):
If the point is to make this posix compliant, this doesn't accomplish it (as the docker check still uses which), but as a step in the right direction seems reasonable

yaronkoren (approved):

install.sh
yaronkoren: These comments are not relevant to the actual code additions, I think - just to the commit summary.

PR #53, opened 2022-12-16
Makefile to build Canasta-CLI

gitmapd:
Makefile:
- Builds the canasta-cli go project.
- Checks syntax errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,40 @@
+APP_NAME="canasta"
+
+build: 
+        @go build -o ${APP_NAME} ./.
+        @go mod tidy
+        @go mod verify
+        @echo "To run the CLI you have just generated, please call it as \033[0;32m./${APP_NAME}\033[0m instead of ${APP_NAME}."
+
+test: 
+        go test -race -cover ./...
+
+benchmark:
+        go test -benchmem -bench . ./...
+
+fmt:
+        test -z $(shell go fmt ./...)
+
+prepare-lint:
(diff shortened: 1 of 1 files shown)


jeffw16:
I want to get a review done by @hexmode first if possible, since he is more experienced with Makefiles.

hexmode:
> I want to get a review done by @hexmode first if possible, since he is more experienced with Makefiles.

I'll try to look at this later this week.

Makefile
```diff
@@ -0,0 +1,40 @@
+APP_NAME="canasta"
+
+build:
```
bawolff, line 3: Should these targets be declared as .PHONY ?

Makefile
```diff
@@ -0,0 +1,40 @@
...
+build: 
+        @go build -o ${APP_NAME} ./.
+        @go mod tidy
+        @go mod verify
+        @echo "To run the CLI you have just generated, please call it as \033[0;32m./${APP_NAME}\033[0m instead of ${APP_NAME}."
+
+test: 
+        go test -race -cover ./...
+
+benchmark:
```
bawolff, line 12: Kind of odd to have a benchmark target when there are no benchmarks.

bawolff (changes requested):
You need to use tabs not spaces. This is not just a stylistic difference in makefiles.

This seems to have a lot of targets that don't do anything (benchmark, test, etc). I suppose they might in future though. I'm not sure if that is wanted.

I think this should  include the extra ldflags like build.sh does.

The makefile seems to unconditionally rebuild, even if no files have changed. I suppose that is intentional?

hexmode:
@bawolff thanks for reviewing this. I lost track of it.

yaronkoren:
An improved version of this patch was implemented in #100.

Issue #25, opened 2022-09-12
Need some sort of progress information during initial create


hexmode:
It just seems to sit there:

``` shellsession
$ canasta create -a WikiSysop -r -w Wiki -n wiki.wiki -i wiki
Setting up Canasta

```

yaronkoren:
I agree that it would be great to have a progress display, with one line getting filled by asterisks or something - I think either yum or apt (or both) do it that way.

aerowisca:
Hello, I've worked on this issue and created a PR for the same. Can one of the maintainers please assign me the issue? Thanks.

hexmode:
@yaronkoren I don't have permission to assign @aerowisca.  Could you?

yaronkoren:
I can assign this (well, anyone could, really - it's just a matter of saying "you are assigned"), but I want to clarify this first.

@aerowisca - first of all, nice job! The patch looks good, and the animated GIFs you added make it very easy to see what it does. (Though the fact that neither command actually worked looks like a major problem, unrelated to your code.)

However, I don't know if you saw, but there is already a patch to create a progress bar, #54. It is not completed, and its developer (@gitmapd) is no longer actively working on it, but it might be the better approach, because it shows a "progress bar" that gets filled in, so users, at least in theory, have a sense for how much longer it will take. Did you see this other PR? Do you have any thoughts about it?

@hexmode - if you have any thoughts on the two approaches, that would be welcome as well.

hexmode:
@yaronkoren initial impressions:
- something that works is better than something that does not
- progress bars can be deceptive since they give you the idea that, say, 50% of the time has past when that may not be the case
- a spinner like the one implemented here looks identical to the one that comes with the glab ort gh cli (I don't remember which atm).

yaronkoren:
These are all valid points - and by "glab ort gh cli" I assume you mean "GitLab or GitHub CLI". I'm guessing it's GitLab, by the way, because I don't remember GitHub having spinners.

To the first point, obviously something that works is better, but my thought was that maybe @aerowisca could switch to using that other patch as the basis for a new patch. On the other hand, as you note, a truly accurate progress bar may be difficult or impossible to accomplish - that's actually one of the main issues with the current progress bar patch.

Anyway, enough delaying: @aerowisca, you are hereby assigned this issue!

[...]

PR #54, opened 2022-12-20
Add progress bar to Canasta Create

gitmapd:


yaronkoren:
Note: this is a fix for issue #25.

cmd/create/create.go
```diff
@@ -68,27 +68,36 @@ func NewCmdCreate() *cobra.Command {
 
 // importCanasta accepts all the keyword arguments and create a installation of the latest Canasta.
 func createCanasta(canastaInfo canasta.CanastaVariables, pwd, path, orchestrator string) error {
+	bar := progressbar.Default(100)
```
bawolff, line 71: This seems to be missing an import

bawolff (changes requested):
When i try to build i get:
```
# github.com/CanastaWiki/Canasta-CLI-Go/cmd/create
cmd/create/create.go:71:9: undefined: progressbar
```

Additionally, I'm not sure this progress bar is meaningfully fine-grained enough. The first time you start Canasta, a lot of the time is going to be spent in the `docker-compose up -d` phase, since all the images have to be downloaded and built, which may be a significant amount of time. I'm not sure that having the progress bar be stuck on 45% is really all that much better than not having a progress bar.

yaronkoren:
A separate approach to this (using spinners instead of a progress bar) was just checked in, in #101.

PR #56, opened 2023-01-02
Add --install and --list-versions flags to install.sh.

gitmapd:
Finds availability for git,jq,docker,docker-compose. Allows to install different versions based on the tags.
Added also when no no flag used it autoinstalls latest version.
Added flag "-l/--list-versions" to list the versions available.
Added flag "-i/--install", it also lists and installs the version we have available.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -45,7 +45,7 @@ containsElement () {
 	return 1
 }
 download_package() {
-	if [[ "$1" =~ ^[0-9]+$ ]] && [[ "$1" < ${#versions[@]} ]]; then
+	if [[ "$1" =~ ^[0-9]$ ]] && [[ "$1" < ${#versions[@]} ]]; then
 			version=${versions[$1]}
 	elif containsElement "$1" "${versions[@]}"; then
 			version=${1:-latest}
--- a/install.sh
+++ b/install.sh
@@ -15,7 +15,7 @@ is_command_available() {
 result=$(is_command_available git docker jq docker-compose)
 
 if [[ $? -ne 0 ]]; then
-    echo "${result} is not installed"
+    echo "${result} is not installed; exiting."
     exit 1
 fi
 
--- a/install.sh
+++ b/install.sh
@@ -13,10 +13,10 @@ else
 fi
 
 loc=$(command -v docker)
-if [ -z $loc ]
+if [ -z "$loc" ]
 then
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -1,44 +1,102 @@
...
-loc=$(which docker-compose)
-if [ -z $loc ]
-then
-    echo "Docker Compose is not installed; please follow the guide at https://docs.docker.com/compose/install/ to install it."
-elif [ -x $loc ]
-then
-    echo "Docker Compose is already installed."
-else
-    echo "Docker Compose appears to be installed at $loc but is not executable; please check permissions."
+result=$(is_command_available git docker jq docker-compose)
```
bawolff, line 15: You don't check for wget and curl, which are both used.

Probably makes sense to pick one of wget or curl, since they both do the same thing and  it would be good to minimize number of dependencies in the install script.

install.sh
```diff
@@ -1,44 +1,102 @@
...
+			else
+				get_versions
+				download_package "$(query_version)"
+			 fi
+			break
+
+                  	;;
+
+		-*)
+			die "Illegal option ${1}"
```
bawolff, line 94: Not critical, but would kind of be nice if this showed a help message listed supported arguments.

install.sh
```diff
@@ -1,44 +1,102 @@
...
+	else
+		canastaURL="https://github.com/CanastaWiki/Canasta-CLI/releases/download/$version/canasta"
+	fi
+
+	wargs=(-q)
+        if wget --help | grep -q -e --show-progress ; then
+		wargs+=(--show-progress)
+	fi
+	echo "Downloading Canasta ${version}"
+	wget "${wargs[@]}" "$canastaURL"
```
bawolff, line 70: Should probably include an argument specifying where to save. If a file named canasta already exists in the directory, this will save it as canasta.2 and then try to move the wrong file.

bawolff (changes requested):
How portable is this shell script meant to be?

If its meant to be super portable so that it can be installed everywhere, I would suggest not targeting bash but targeting /bin/sh, avoiding using jq (Commonly not installed on servers), and maybe picking one of either wget or curl to use.

[My main request change is for not including wget and curl in the list of binaries to check for. My other comments are more nice to haves]

yaronkoren:
Closing this PR - this functionality, more or less, was added in with a separate patch, #77.

PR #70, opened 2023-02-25
Add github actions workflow to build and release

### LINKED ISSUE #68: Releases aren't updated often
Currently, executables in releases are updated manually. This leads to less frequent updates, and also when using the `install.sh` script the users doesn't get the absolute latest updates made to the repository. This could result in issues due to missing bug fixes.

The proposed solution is CI/CD pipeline to build and update the executables on every push/pull request made to the `main` branch. So there would be a new release on each update to the main branch.  

Should there be any change to the frequency of releases?

amalpaul54111:
Fixes #68


amalpaul54111:
We would have to create a new `releases` branch and on each push to this branch, a new release would be automatically created. Currently, no changelog is being generated. If we think frequent releases are okay, we could just make use of the `main` branch to create releases.

jeffw16 (approved):
LGTM

PR #71, opened 2023-02-25
Change default directory permissions to be more restrictive (#46)

### LINKED ISSUE #64: Canasta config strategy
Filing as a follow up to #48 

Canasta needs some persistent state to keep track of wikis it manages. Right now it stores that state in /etc/canasta if you are root, and ~/.config/canasta if you are not root. If these directories/files do not exist, then the directory is created 0777 and the file is created 0644.

I think first of all, that it makes more sense to create the directory as 0755 not 0777 since it is a per-user directory. Normally I think of /etc as a place for global configuration files that are not generally written to by the application and apply to all users. I think it might make more sense to have the root state file live in /root/.config like the other users.

I think there is the larger question of when castana is used by different users, should they share information on what wikis exist, or should every user get a different view. I'm not sure what the correct answer to that is. If it is expected that different users all have the same view of which wikis are managed by castana, then i would expect the file to live in somewhere like /var/lib or /var/local.

[Edit: As i write this out more fully, i realize that maybe all this doesn't really matter all that much]
[Edit: fix typo]

bawolff:
Make the config directory be 0755. Make the restic backups directory be 0700.

Fixes #64

### DIFF
--- a/cmd/restic/restic.go
+++ b/cmd/restic/restic.go
@@ -64,7 +64,7 @@ func checkCurrentSnapshotFolder(currentSnapshotFolder string) {
 	if _, err := os.Stat(currentSnapshotFolder); err != nil {
 		if os.IsNotExist(err) {
 			logging.Print("Creating..." + currentSnapshotFolder)
-			if err := os.Mkdir(currentSnapshotFolder, os.ModePerm); err != nil {
+			if err := os.Mkdir(currentSnapshotFolder, 0o700); err != nil {
 				logging.Fatal(err)
 			}
 		} else {
@@ -75,7 +75,7 @@ func checkCurrentSnapshotFolder(currentSnapshotFolder string) {
 		if err := os.RemoveAll(currentSnapshotFolder); err != nil {
 			logging.Fatal(err)
 		}
-		if err := os.Mkdir(currentSnapshotFolder, os.ModePerm); err != nil {
+		if err := os.Mkdir(currentSnapshotFolder, 0o700); err != nil {
 			logging.Fatal(err)
 		}
 		logging.Print("Emptied.. " + currentSnapshotFolder)
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -123,7 +123,7 @@ func GetConfigDir() string {
 	fi, err := os.Stat(dir)
 	if os.IsNotExist(err) {
 		log.Print(fmt.Sprintf("Creating %s\n", dir))
-		err = os.MkdirAll(dir, os.ModePerm)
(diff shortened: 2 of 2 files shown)


cmd/restic/restic.go
```diff
@@ -64,7 +64,7 @@ func checkCurrentSnapshotFolder(currentSnapshotFolder string) {
 	if _, err := os.Stat(currentSnapshotFolder); err != nil {
 		if os.IsNotExist(err) {
 			logging.Print("Creating..." + currentSnapshotFolder)
-			if err := os.Mkdir(currentSnapshotFolder, os.ModePerm); err != nil {
+			if err := os.Mkdir(currentSnapshotFolder, 0o700); err != nil {
```
jeffw16, line 67: What if the user doesn't want their backups to have permissions that are this restrictive?

naresh-kumar-babu:
@bawolff  '0o700' should change to '0700' and '0o755' to '0755'.

bawolff:
> @bawolff '0o700' should change to '0700' and '0o755' to '0755'.

Why?

I mean, i dont really care either way, but is there a reason you think one is better than the other?

naresh-kumar-babu:
Yeah, this is the common format for permission modes

jeffw16:
Naresh, that might be true, but is there a reason `0o755` is flat out wrong? Otherwise, I don't see what the problem is. It's personal preference. Unless you think that this is confusing to the majority of people (which I can see being a problem), we shouldn't point out issues that are not actually issues.

bawolff:
For context, 0755 is traditional because 0o755 is much newer format. Some people feel that it is very easy to misread 0755 and not see the leading 0, where 0o is much more obvious what is going on. I don't have a strong preference either way

naresh-kumar-babu:
@bawolff I didn't mean that it was wrong. It's just that we are used to 0755 a lot. It's more known than the latter. But as you said, it does not bring any functional advantages. It's up to you.

jeffw16 (commented):
Forgot to send feedback I wrote 11 months ago

cmd/restic/restic.go
bawolff: I guess its impossible to account for all possible use cases. Would 0770 be better to allow for there to be a group of users who can access it?

These backups do have user passwords in them, so they are kind of sensitive

jeffw16: Hmm understandable. Well, please don't let my comment be a blocker. I'll approve this for now, as I don't remember my exact logic from 11 months ago.

jeffw16 (approved):

PR #75, opened 2023-03-21
Provided Real-Time Output for External Subcommands

### LINKED ISSUE #73: Enhancement for Issue #61: Provide Real-Time Output for External Subcommands
Issue #61 brings attention to the problem of insufficient information when running external subcommands, such as docker-compose. While it may seem that passing the verbose flag to subcommands could resolve this issue, I believe it's not enough. This is because our Run() function currently only prints the output of external commands after they have finished executing.

Thus, even if we pass verbose to the external subcommand, we would only receive the information after it has completed its work. 

To address this limitation, I propose that we modify the Run function to provide users with real-time output from external commands. This way, they can have all the information they need to understand what is happening during long-running commands as they happen, instead of having to wait until the command is finished.

chl178:
Fixes #73

This updated version of the execute package adds a custom writerWithPrint struct that implements the io.Writer interface. This allows the output of the executed command to be printed to the console while also being captured in a buffer.

The writerWithPrint struct has a buf field of type bytes.Buffer. The Write method of the struct uses logging.Print function to print and writes the bytes to the buffer. So when we have verbose, it will provide the output of external commands in real time.

## Test

I have tested all external subcommands including docker-compose, git, etc. When we have verbose, the new 'Run' function is able to successfully provide their real-time output. If there are bugs during their execution, the function will print the output of external subcommands we captured as before.

Please review and let me know if you have any questions or concerns.

Thank you.


jeffw16:
Best of luck with your finals! Please don't reply until you have time.

May I ask why we are using a multithreaded implementation with locks here? It seems like overkill for the purposes of the CLI (although I think it might help with all of the logging happening within the Canasta image). Locks take time to lock and unlock. If we are running a single-threaded process, there is no point in locking our writes to stdout.

jeffw16:
Also, forgive me if it turns out that `MultiWriter` is the de facto generic implementation of the `io.Writer` interface. I don't write Go, nor am I a huge fan of the language.

chl178:
@jeffw16 Thank you for your suggestion, you were correct. In a single-threaded process, there is no need to lock.

chl178:
@jeffw16 @yaronkoren @amalpaul54111 would you please take a look at this?

jeffw16:
The code itself looks good, but I don't have time to try it just yet.

yaronkoren (approved):

yaronkoren (approved):

PR #91, opened 2023-07-23
Add support for copying override file on create/import

### LINKED ISSUE #90: Allow a docker-compose.override.yml to be specified for create and import
As far as I can tell, there is currently no way to provide a docker-compose.override.yml file until after the containers have already been created. It would be good to have a parameter to the create and import commands that would allow an override file to be specified before the containers are created.

cicalese:
This should fix #90.

### DIFF
--- a/cmd/create/create.go
+++ b/cmd/create/create.go
@@ -64,7 +64,7 @@ func NewCmdCreate() *cobra.Command {
 	createCmd.Flags().StringVarP(&canastaInfo.AdminName, "WikiSysop", "a", "", "Initial wiki admin username")
 	createCmd.Flags().StringVarP(&canastaInfo.AdminPassword, "password", "s", "", "Initial wiki admin password")
 	createCmd.Flags().BoolVarP(&keepConfig, "keep-config", "k", false, "Keep the config files on installation failure")
-	createCmd.Flags().StringVarP(&override, "override", "r", "", "Source filename for override file to copy")
+	createCmd.Flags().StringVarP(&override, "override", "r", "", "Name of a file to copy to docker-compose.override.yml")
 	return createCmd
 }
 
--- a/cmd/create/create.go
+++ b/cmd/create/create.go
@@ -16,13 +16,13 @@ import (
 
 func NewCmdCreate() *cobra.Command {
 	var (
-		path             string
-		orchestrator     string
-		pwd              string
-		err              error
-		keepConfig       bool
-		canastaInfo      canasta.CanastaVariables
-		overrideFilename string
+		path         string
+		orchestrator string
+		pwd          string
+		err          error
+		keepConfig   bool
+		canastaInfo  canasta.CanastaVariables
+		override     bool
 	)
 	createCmd := &cobra.Command{
 		Use:   "create",
@@ -33,7 +33,7 @@ func NewCmdCreate() *cobra.Command {
 				log.Fatal(err)
 			}
 			fmt.Println("Creating Canasta installation '" + canastaInfo.Id + "'...")
-			if err = createCanasta(canastaInfo, pwd, path, orchestrator, overrideFilename); err != nil {
+			if err = createCanasta(canastaInfo, pwd, path, orchestrator, override); err != nil {
 				fmt.Print(err.Error(), "\n")
 				if keepConfig {
 					log.Fatal(fmt.Errorf("Keeping all the containers and config files\nExiting"))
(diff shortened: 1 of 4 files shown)


cicalese:
I'm not sure how to test this locally.

hexmode:
Build error:
``` console
$ go build
go: downloading github.com/spf13/cobra v1.4.0
go: downloading github.com/kirsle/configdir v0.0.0-20170128060238-e45d2f54772f
go: downloading github.com/sethvargo/go-password v0.2.0
go: downloading golang.org/x/term v0.0.0-20220526004731-065cf7ba2467
go: downloading golang.org/x/sys v0.0.0-20210615035016-665e8c7367d1
# github.com/CanastaWiki/Canasta-CLI-Go/cmd/import
cmd/import/importExisting.go:69:2: undefined: createCmd

**Comment by jeffw16** [comment]
My impression of this PR is: it seems to allow the user to copy over an override file that they can name anything (including names other than `docker-compose.override.yml`). Technically the rest of the CLI is not able to handle that yet.

Also, the concept of an override file seems to be pretty specific to Docker Compose, so this flag should probably fail when invoked for Kubernetes (i.e. when `-o kubernetes` is passed in rather than `-o docker-compose`).

**Comment by cicalese** [comment]
Yes, I was unsure what to do about other orchestrators. As far as I can tell, the CLI doesn't have a way to add additional arguments on a per-orchestrator basis. That's why I made it so that you can actually copy any single file with this argument, not just docker-compose.override.yml. I'm not 100% happy with it, either, but it is not tied only to docker-compose that way.

**Comment by cicalese** [comment]
I have tested this locally and confirmed that it works as designed and addresses the issue we needed it for.

**Comment by jeffw16** [comment]
What are y'all naming the override file? Are you just passing in `docker-compose.override.yml`?

**Comment by cicalese** [comment]
Yes, I am indeed just passing in `docker-compose.override.yml`.

File: cmd/create/create.go [hunk 1]
```diff
@@ -32,7 +33,7 @@ func NewCmdCreate() *cobra.Command {
 				log.Fatal(err)
 			}
 			fmt.Println("Creating Canasta installation '" + canastaInfo.Id + "'...")
-			if err = createCanasta(canastaInfo, pwd, path, orchestrator); err != nil {
+			if err = createCanasta(canastaInfo, pwd, path, orchestrator, overrideFilename); err != nil {
```
jeffw16: The Canasta CLI's support for Docker Compose is currently not designed to use an override file that is named anything other than `docker-compose.override.yml`. Would it be possible to change this to a binary flag and eliminate the ability to pick the filename?

jeffw16 (changes requested):

cmd/create/create.go
cicalese: > Would it be possible to change this to a binary flag and eliminate the ability to pick the filename?

Yes, but that would hardcode this feature to work only with docker-compose. Another option would be to create a way to pass parameters to the orchestrator on creation, but that would be a much larger architectural change. Another option would be to change it to copy one or more arbitrary files, to be used however the orchestrator or end user sees fit. That would essentially be the same functionality with slightly broader semantics.

jeffw16 (approved):
LGTM

cicalese:
It turned out to be more straightforward than I had anticipated to be able to have the parameter only work when the orchestrator is docker-compose.

This is tested and working.

I do still have a slight preference to providing the filename rather than using a boolean, so that you could specify different override files for different situations (e.g. different port values, different images, replacing some services) . Regardless of the input file name, the file would be copied to docker-compose.override.yml if docker-compose is the orchestrator. If you agree, I can make that change. Otherwise, this should be good to go.

jeffw16:
Great, thank you @cicalese!

While I also would prefer to give users the option, the other CLI commands currently don't support using override files that aren't supported by default. For instance, when one does `canasta start`, it calls `docker compose up -d` but it doesn't pass in any other manifest files. If the support is added to all commands, then I think this would be a great feature to add to the Canasta CLI.

cicalese:
Actually, the create and import commands allow you to specify a .env file, and the import command allows you to specify a LocalSettings.php file, which is what I initially modeled this after.

As for the other commands, it makes sense (to me) that the override file would be specified when the container is created (or imported, which is a type of create) but not for the start/stop commands. If you could specify an override for the later commands, then what would be the meaning if you do not supply an override file when you start but you did when you created? Would you delete the override file? Besides, after the initial create/import, the directory has been created, and the override file can be added/edited/deleted at will. The reason you need it for create/import is that there is no other way to specify it at that point.

cicalese:
I'm going to add the filename bit in so you can see that it is a minimal change and still is tied only to the docker-compose orchestrator. That way I can also test the patch locally. If you prefer the boolean to the filename after you see it, we can back the commit back out.

cicalese:
OK, this is tested and confirmed to be working as desired.

To reiterate, the reason this is necessary for the create and import commands is that they are run to create the directory that would hold the canasta instance, including the override file, and then they start the containers. The override file is needed before the containers are started, but it cannot be placed in the canasta directory before the commands are run since the directory does not exist yet at that point.

jeffw16 (approved):
Ok, this LGTM

yaronkoren (approved):

PR #95, opened 2023-08-13
Only enable Vector skin by default

### LINKED ISSUE #85: Remove automatically-added wfLoadSkin() calls from LocalSettings.php
While moving wikiapiary, we noticed all the enabled skins were enabled in the embedded LocalSettings.php and could not (easily) be disabled.  The embedded LocalSettings should include as little customization possible and, where possible, it should be modifiable.

cicalese:
Partially fixes #85 

This patch makes it so the installer only enables the Vector skin, not all of the other installed skins. It isn't possible to get the installer to enable zero skins, but this is at least a bit of an improvement.

### DIFF
--- a/internal/mediawiki/mediawiki.go
+++ b/internal/mediawiki/mediawiki.go
@@ -63,7 +63,7 @@ func Install(path, orchestrator string, canastaInfo canasta.CanastaVariables) (c
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


yaronkoren:
This is great! I didn't know there was a "skins" flag. This looks good to me.

yaronkoren (approved):

PR #99, opened 2024-01-14
Allow root DB password and wiki DB username and password to be changed

cicalese:
Currently the root DB password is hardcoded to "mediawiki" and the wiki DB user is hardcoded to be the root DB user. This patch allows the root DB password to be set or generated. It allows a separate DB user to be set as the wiki user, also with a password that can be set or generated. As is currently done with the wiki admin password, if the user does not enter a password when prompted, a password is generated and saved to a file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,9 @@ sudo canasta create [flags]
 - `-f, --yamlfile`: Initial wiki yaml file for wiki farm setup.
 - `-k, --keep-config`: Keep the config files on installation failure.
 - `-r, --override`: Name of a file to copy to docker-compose.override.yml.
-- `--rootdbpass`: Prompt for the password for the root database user (default: "mediawiki").
-- `--wikidbuser`: The database user to use for normal operations (default: "root").
-- `--wikidbpass`: Prompt for the password for the database user to use for normal operations (default: "mediawiki").
+- `--rootdbpass`: Read root database user password from .root-db-password file or prompt for it if file does not exist (default password: "mediawiki").
+- `--wikidbuser`: The username of the wiki database user (default: "root").
+- `--wikidbpass`: Read wiki database user password from .wiki-db-password file or prompt for it if file does not exist (default password: "mediawiki").
 
 **YAML Format for Wiki Farm:**
 To create a wiki farm, you first need to create a YAML file with the following format:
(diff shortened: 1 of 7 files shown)


yaronkoren:
This generally looks great - but what is the need for the .root-db-password and .wiki-db-password files? I guess really my question is: are there other command-line tools that allow for this kind of password-file approach? I haven't heard of such a thing.

cicalese:
> what is the need for the .root-db-password and .wiki-db-password files?

If the passwords are auto-generated, you need a way to know what they were set to. Arguably, you could go digging around in the db and web containers to find the values. They are also necessary to pass to the add wiki CLI command, so digging around in the containers would not be feasible for that.

> I guess really my question is: are there other command-line tools that allow for this kind of password-file approach? I haven't heard of such a thing.

I followed the style already in the code for the wiki admin password. In that case, the user is prompted for the password. If they just hit Enter, a password is auto-generated and stored in a file. I chose to take the same approach for these two passwords, because it seemed it would be odd to do something different than what is already present in the code.

yaronkoren:
Okay - I knew that Canasta will save an auto-generated password to a file, but I didn't know that it can read from that same file. Does it? Or are there different circumstances for this set of passwords? Apologies for asking you to explain things I should probably know about Canasta already. :)

cicalese:
> Okay - I knew that Canasta will save an auto-generated password to a file, but I didn't know that it can read from that same file. Does it?

Yes, when it is adding a wiki, it reads the admin password from the file so it can set the admin password on the new wiki to the same string it is set to on the original wiki: https://github.com/CanastaWiki/Canasta-CLI/blob/main/internal/mediawiki/mediawiki.go#L145

README.md
freephile, line 1: I suggest that in this README (and possibly elsewhere), where we describe that an administrator can provide a DB User and DB password, that we link to the documentation on mediawiki.org about the requirements for said user: namely that it needs ALL_PRIVILEGES. https://www.mediawiki.org/wiki/Manual:Installing_MediaWiki#Set_up_the_database

freephile: The new DB user is created BY the MediaWiki installer with appropriate permissions (GRANTS)!

freephile (approved):
Looks great, THANKS

PR #103, opened 2024-03-13
Fix wiki name directory issue while performing create action

naresh-kumar-babu:
- Replace spaces to underlines.
- Remove accented and non-alphanumeric characters

### DIFF
--- a/cmd/create/create.go
+++ b/cmd/create/create.go
@@ -5,8 +5,6 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"strings"
-	"regexp"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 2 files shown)


cmd/create/create.go
```diff
@@ -40,6 +42,11 @@ func NewCmdCreate() *cobra.Command {
 			if name, canastaInfo, err = prompt.PromptUser(name, yamlPath, rootdbpass, wikidbpass, canastaInfo); err != nil {
 				log.Fatal(err)
 			}
+			
+			// Replace spaces to underlines and remove accented and non-alphanumeric characters
+			name = strings.Replace(name, " ", "_", -1)
```
jeffw16: What does the name refer to? Directory name or the wiki name? It does not make sense to have `$wgSitename` be something like "My_New_Canasta_Wiki".

naresh-kumar-babu: It is the Wiki name.

jeffw16: In that case it would be silly to replace the spaces with underscores.

jeffw16 (changes requested):
I have concerns about replacing the spaces with underscores for $wgSitename

yaronkoren (approved):

yaronkoren:
@jeffw16 - please re-review! The patch is totally different now.

PR #104, opened 2024-04-16
Move password length validation to pre-create stage

### LINKED ISSUE #373: New wiki setup wizard accepts password less than 10 characters
Steps to reproduce
1. Follow the "Installation" and "Create new wiki" steps at http://canasta.wiki/setup/ for CLI installation
2. Enter a password less than 10 characters when prompted

What happens: The wizard accepts the input, attempts the installation, and then fails due to the lack of a 10-character password

What should happen: The wizard ideally would reject invalid passwords (at the very least, passwords <10 characters) and not proceed until a valid password is entered.

naresh-kumar-babu:
Fixes https://github.com/CanastaWiki/Canasta/issues/373

### DIFF
--- a/internal/prompt/prompt.go
+++ b/internal/prompt/prompt.go
@@ -111,6 +111,15 @@ func promptForUserPassword(username, password string) (string, string, error) {
 	return username, password, nil
 }
 
+func passwordCheck(username string, password string) (error) {
+	if len(password) < 10 {
+		return fmt.Errorf("Password must be at least 10 characters long ")
+	} else if strings.Contains(password, username) || strings.Contains(username, password) {
+		return fmt.Errorf("Password should not be similar to the username")
+	}
+	return nil
+}
+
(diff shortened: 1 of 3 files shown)


internal/prompt/prompt.go
```diff
@@ -125,6 +125,11 @@ func getAndConfirmPassword(username string) (string, string, error) {
 	if err != nil || password != confirmedPassword {
 		return "", "", fmt.Errorf("Passwords do not match, please try again.")
 	}
+	if len(password) < 10 {
```
jeffw16: Why did we move this away from its own function? This makes it less unit testable.

jeffw16 (changes requested):

jeffw16:
Can we write some unit tests to ensure this function works correctly?

naresh-kumar-babu:
> Can we write some unit tests to ensure this function works correctly?

Added unit test.

jeffw16 (approved):

PR #117, opened 2024-09-09
Throw error in case of space or non-ascii character in CanastaID

naresh-kumar-babu:
- Throw error in case of space or non-ascii character in CanastaID.


yaronkoren (approved):
Looks great!

PR #118, opened 2024-09-10
Allow dashes and underscores in canastaID

naresh-kumar-babu:
- Allow dashes and underscores in canastaID, except as a first or last character.


yaronkoren (approved):

labster:
Generally if you want dash in a character class, you put it as the very first character in that class to avoid confusion.

PR #129, opened 2025-10-28
Add support for AArch64 builds

jeffw16:
The only issue is that the install script will need to be changed to not just use an executable named `canasta`. It will be a different file name depending on the architecture.


yaronkoren:
Is it possible to always have the executable be named "canasta", regardless of the architecture? Otherwise, that's a pretty major change.

jeffw16:
Closed in favor of #135

PR #134, opened 2026-01-01
Fix chmod permissions issue in install.sh

### LINKED ISSUE #132: Canasta-CLI  /install.sh doesn't catch fail on chmod
When running Canasta-CLI install.sh, if chmod in l147 fails the installer still claims success. 
This failed even after running the curl as sudo.

buterfloro:
`chmod` in line 147 can fail when the install script isn't running as root because it's trying to change a file owned by a different user, the root user. 

Running `chmod` with `sudo` fixes this issue, as `sudo` allows `chmod` to run as root, the owner of the file `chmod` is changing.

Tested working on a fresh Kubuntu 25.10 livecd virtual machine with distribution-packaged Docker. 

Fixes: #132

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -144,7 +144,7 @@ download_and_install() {
     exit 1
   fi
   echo "Download was successful; now installing Canasta CLI."
-  chmod u=rwx,g=xr,o=x canasta
+  sudo chmod u=rwx,g=xr,o=x canasta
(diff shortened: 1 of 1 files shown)


cicalese (approved):
Looks good to me.

yaronkoren:
@buterfloro - thank you for this fix!

PR #135, opened 2026-01-04
Add support for building for MacOS

cicalese:


cicalese:
I didn't notice #129 until I was ready to push this. This PR supports MacOS in addition to supporting multiple architectures.

cicalese:
Note that, while it does generate 4 cross-compiled executables, the install script chooses the correct one and downloads it to 'canasta' locally.

Also note that this has been tested on MacOS, where it works as long as you do not run the 'canasta' commands with sudo. You still use sudo for 'install.sh' so that the executable can be saved to the appropriate directory, but the canasta commands are run as an unprivileged user. If you run with sudo, files are created by canasta with root ownership that cause later failures.

jeffw16:
I’m not entirely familiar with GitHub Actions, but just wanted to check to be on the safe side, is there a need to specify the specific type of architecture and platform it’s being built for, before the image is actually built, like with a platform matrix? Am just unfamiliar with this approach since #129 takes a different approach. Nonetheless, this looks good to me as long as it’s working as intended.

jeffw16 (approved):

cicalese:
Yeah, after I saw your patch, I realized that it could have used a matrix and called the build script 4 times rather than having the platform logic in the build script. I think that using a matrix might be a bit cleaner, so I'll investigate that option.

cicalese:
OK, I've updated it to generate a single executable in build.sh and to use a matrix in the GitHub action.

cicalese:
I also made the install script backward compatible to the old executable name.

jeffw16 (approved):

build.sh
```diff
@@ -2,6 +2,31 @@
...
+# Usage:
+#   ./build.sh                           # Build for current platform
+#   GOOS=linux GOARCH=amd64 ./build.sh   # Cross-compile for specific platform
 
-go build -ldflags="-X 'github.com/CanastaWiki/Canasta-CLI-Go/cmd/version.sha1=$(git rev-parse --short HEAD)'
-    -X 'github.com/CanastaWiki/Canasta-CLI-Go/cmd/version.buildTime=$(date +'%Y-%m-%d %T')'"
+# Get version information
+GIT_SHA=$(git rev-parse --short HEAD)
+BUILD_TIME=$(date +'%Y-%m-%d %T')
+LDFLAGS="-X 'github.com/CanastaWiki/Canasta-CLI-Go/cmd/version.sha1=${GIT_SHA}' -X 'github.com/CanastaWiki/Canasta-CLI-Go/cmd/version.buildTime=${BUILD_TIME}'"
```
yaronkoren: Shouldn't the "-Go" be removed here?

install.sh
```diff
@@ -125,23 +125,78 @@ check_wget_show_progress() {
...
+  case "$(uname -m)" in
+    x86_64|amd64)
+      arch="amd64"
+      ;;
+    aarch64|arm64)
+      arch="arm64"
+      ;;
+    *)
+      echo "Unsupported architecture: $(uname -m)"
+      echo "Canasta CLI supports amd64 and arm64 only."
```
yaronkoren: Should this also mention x86 and aarch64, or would that be unnecessary for most people? I don't know the answer to this.

jeffw16: amd64 == x86-64
arm64 == AArch64

yaronkoren (commented):
This looks great! it's great to now be able to support different builds for different architectures. I have two minor comments about the code that I put in - let me know what you think.

build.sh
cicalese: Yes, absolutely. Thanks for catching that.

yaronkoren:
Great! What do you think about the "supports amd64 and arm64 only" wording, though? Good enough? I personally am much more familiar with the term "x86" than AMD, for example, but maybe most people understand all of them.

yaronkoren (approved):
Thanks, it looks great now!

cicalese:
I was mulling making the same clarification in the README, but that can be done later if we feel it is necessary. Thanks for the merge!

yaronkoren:
Oh, yeah. If we're getting into minor details, I try to avoid capitalizing all words, so for example I think "Supported Platforms" in the README would be better as "Supported platforms". Too minor to note for this patch, but, if we're already talking about stuff like that.

cicalese:
I have confirmed that running

```
curl -fsL https://raw.githubusercontent.com/CanastaWiki/Canasta-CLI/main/install.sh | bash
```

on a Mac now works correctly.

yaronkoren:
Awesome. 👍

PR #139, opened 2026-01-13
Fix broken add command, remove prompts, and refactor password handling

cicalese:
Fixes broken add command that was ordering operations incorrectly to
leave the wiki farm in an inconsistent state. Makes create and add CLI
commands fully scriptable by removing interactive prompts. Implements
consistent password handling across create, add, and import commands.

### Breaking Changes
- create: rename --WikiSysop to --admin (breaking change)
- create: change --rootdbpass and --wikidbpass from boolean to string
  flags (breaking change - behavior now: flag value or auto-generate
  instead of prompt)
- add: move --site-name shortcut from -s to -t to free up -s for
  --password (breaking change for users using -s shortcut)

### New Flags
- create: add --envfile (-e) to merge custom env file with .env.example
- add: add --password (-s) flag for admin password override

### Required Flags
- create: make --id and --admin required
- add: make --wiki, --url, --id, and --admin required

### Consistency Improvements
- Unify flag shortcuts and descriptions across commands

## Terminology Clarification

Establish clear, consistent terminology throughout codebase:
- **Canasta instance ID**: identifies the wiki farm installation (which only
  contains a single wiki after "canasta create" or multiple wikis if "canasta add"
  was also invoked)
- **Wiki ID**: unique identifier for a specific wiki (used in config,
  database names, directories, filenames)
- **Site name**: display name shown to users (stored in wikis.yaml 'name'
  field, used in $wgSitename)

### Variable and Function Renaming
- Rename all `wikiName` variables to `wikiID`
- Rename all function parameters from `name` to `id` for wiki identifiers
  (in farmsettings, mediawiki, canasta packages)
- Rename `WikiNames` array variables to `WikiIDs`
- Rename `WikiNameExists()` to `WikiIDExists()`
- Rename `CheckRunningStatus` parameter from ambiguous `id` to `canastaID`
- Add `ReadWikisYamlWithNames()` helper to read wiki metadata with site names
- Use `canastaID` consistently for Canasta instance identifier

### Help Text and Messages
- Update all flag help text to use correct terminology:
  * "Canasta instance ID" for farm-level operations
  * "ID of the wiki" for wiki-specific operations
  * "Display name of the wiki" for site name
- Update list command table header from "Wiki ID(Name)" to "Wiki ID"

### Documentation
- Update README.md with corrected flag names and descriptions

## Password Handling Changes

- Admin passwords: use --password flag value if provided, otherwise
  auto-generate and save to config/admin-password_{wikiid}
- DB passwords: use --rootdbpass/--wikidbpass flag values if provided,
  otherwise auto-generate and save to .env only
- Add --envfile flag to create command to merge custom env files with
  .env.example (allows providing passwords via file instead of flags)
- Change --rootdbpass and --wikidbpass from boolean prompt triggers to
  string flags accepting password values
- Remove all password file reading from working directory dot files
- Database credentials now stored exclusively in .env (single source)
- Admin password files moved from working directory to installation
  config/ directory with non-hidden filenames

## Code Structure Improvements

- Delete internal/prompt package (no longer needed without prompts)
- Move ValidateWikiID from prompt package to farmsettings package
- Split CopyEnv into CreateEnvFile (for create) and CopyEnvFile (for
  import) to clarify different use cases
- Split CheckWiki into WikiIDExists and WikiUrlExists for better
  separation of concerns
- Refactor GeneratePasswords to handle all password types
- Add GeneratePassword and SavePasswordToFile helper functions
- Update RewriteSettings to properly use site name for $wgSitename
  (display name) and wiki ID for $wgMetaNamespace (identifier)

## Bug Fixes

- Fix CopySetting to actually copy Settings.php (was creating directory
  but not copying file) - this was breaking add command
- Fix add command to create per-wiki Settings.php like create does
- Fix add command to properly order operations (install before adding to
  wikis.yaml) to prevent inconsistent state on failure
- Fix create and import command to not prompt about keeping files on error (leaves
  command line option to trigger this behavior)
- Fix CheckRunningStatus to use canastaID parameter in error messages

## Build System

- Update Makefile to call build.sh for consistent versioned builds
- Copy build/canasta-{GOOS}-{GOARCH} to /usr/local/bin/canasta
- Add clean target to remove build artifacts
- Unify build output location in build/ directory


cicalese:
This is a *much* bigger PR than I would normally create. It fixes several interconnected issues that were pervasive across the codebase. I have tested it locally (on a Mac), and I am able to create a wiki farm, add wikis, delete wikis, and remove the wiki farm.

internal/farmsettings/farmsettings.go
```diff
@@ -20,10 +20,29 @@ type Wikis struct {
 	Wikis []Wiki `yaml:"wikis"`
 }
 
-func CreateYaml(name, domain string, path *string) error {
+// ValidateWikiID validates that wikiID doesn't contain invalid characters or reserved names
+func ValidateWikiID(wikiID string) error {
```
jeffw16, line 24: Are there plans to add unit tests for this function?

cmd/add/add.go
```diff
@@ -13,31 +15,94 @@ import (
...
+			if err := farmsettings.ValidateWikiID(wikiID); err != nil {
 				log.Fatal(err)
 			}
-			fmt.Printf("Adding wiki '%s' to Canasta instance '%s'...\n", wikiName, instance.Id)
-			err = AddWiki(wikiName, domainName, wikiPath, siteName, databasePath, admin, wikidbuser, instance)
+
+			// Parse URL to extract domain and path
+			urlString := url
+			if !strings.HasPrefix(urlString, "http://") && !strings.HasPrefix(urlString, "https://") {
+				urlString = "https://" + urlString
```
jeffw16, line 86: Do we want to assume `http://` for `localhost` or `127.0.0.1`?

cmd/add/add.go
```diff
@@ -132,7 +206,7 @@ func AddWiki(name, domain, wikipath, siteName, databasePath, admin, wikidbuser s
 		return err
 	}
 
-	fmt.Println("Successfully Added the Wiki '" + name + "in Canasta instance '" + instance.Id + "'...")
+	fmt.Println("Successfully added wiki '" + wikiID + "' in Canasta instance '" + instance.Id + "'...")
```
jeffw16, line 209: Nit: in Go, the prevailing style does not have spaces between `+` and strings in concatenation.

internal/config/config.go
```diff
@@ -103,21 +103,21 @@ func ListAll() {
...
+func GetCanastaID(installPath string) (string, error) {
+	var canastaID string
 	for _, installations := range existingInstallations.Installations {
-		if installations.Path == path {
+		if installations.Path == installPath {
 			return installations.Id, nil
 		}
 	}
-	return canastaId, fmt.Errorf("No canasta installations exist at %s", path)
+	return canastaID, fmt.Errorf("No canasta installations exist at %s", installPath)
```
jeffw16: Nit: canasta -> Canasta

internal/farmsettings/farmsettings.go
cicalese: Right now there are no tests in the codebase. There was a single test for some of the prompt functionality, but this PR removes it since the functionality it was testing was removed. I have some work in progress to begin adding tests. CLI absolutely needs a comprehensive set of tests, but that will involve some refactoring and will likely be another large PR.

cmd/add/add.go
cicalese: Really? I did a quick search and did not find an example of that. I'm not averse to changing that if it is the case, but since there are many instances of this in the code, it would be better to address style issues in a separate PR. This one is big enough already. It looks like `gofmt` would be a good choice to address code style.

cmd/add/add.go
cicalese: Unclear. This is the existing code that was previously in `PromptWiki` in prompt.go. Since it currently works, we should address any changes in behavior in a separate PR.

internal/config/config.go
cicalese: Done.

naresh-kumar-babu (approved):
Tested - `create`  and `add` commands - both working as expected.

jeffw16 (approved):

yaronkoren (approved):

PR #147, opened 2026-01-28
Make --id flag optional when running from an instance directory

cicalese:
All commands that operate on an existing Canasta instance now resolve the instance from the current working directory when --id is not provided. This uses the existing CheckCanastaId() fallback which looks up the instance path in the config registry.

Commands that also accept the instance ID as a positional argument (start, stop, restart, upgrade, delete) retain that behavior.

The create command is unchanged since it always requires --id to name the new installation.

How it works:

  1. The change sets instance.Path to the working directory.  
  2. CheckCanastaId() (internal/canasta/canasta.go:524) checks: if instance.Id  
  is empty (no -i flag), it calls config.GetCanastaID(instance.Path) at line 531
  3. GetCanastaID() (internal/config/config.go:116) iterates through all        
  installations stored in conf.json and matches by path:                        
```
  for _, installations := range existingInstallations.Installations {           
      if installations.Path == installPath {                                    
          return installations.Id, nil                                          
      }                                                                         
  }                                                                             

```
So it looks up the working directory in the config registry (.../canasta/conf.json), which stores the path for every installation created with canasta create. If your current directory matches a registered installation path, it returns that installation's ID.

Before #146 was merged, all these commands had a --path flag that defaulted to the working directory:                          
                                                                                
startCmd.Flags().StringVarP(&instance.Path, "path", "p", workingDir, "Canasta installation directory")                                                      
                                                                                
That flag's default value silently set instance.Path to the working directory, which made CheckCanastaId() work without -i. When that PR removed --path, it inadvertently broke the path-based lookup because nothing was setting instance.Path anymore.This change restores that behavior by setting instance.Path = workingDir directly, without re-introducing the --path flag.

### DIFF
--- a/cmd/add/add.go
+++ b/cmd/add/add.go
@@ -64,6 +64,12 @@ func NewCmdCreate() *cobra.Command {
 	var adminPassword string
 	var wikidbuser string
 
+	workingDir, err := os.Getwd()
+	if err != nil {
+		log.Fatal(err)
+	}
+	instance.Path = workingDir
+
 	addCmd := &cobra.Command{
 		Use:   "add",
 		Short: "Add a new wiki to a Canasta instance",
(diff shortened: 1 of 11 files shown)


yaronkoren:
This definitely can simplify things for admins... but could this be "too clever"? I just don't know of a case where a command makes use of one's current directory for context - other than, of course, for stuff like "ls".

cicalese:
It was already the case for some, but not all, of the commands. This just makes it consistent and fixes the breakage from the previous PR. I actually use it all the time and was annoyed at the few commands it didn't work for. You can always override it by adding the instance flag.

yaronkoren:
Oh! Never mind, then.

yaronkoren (approved):

PR #149, opened 2026-01-28
Fix import command, add export, add import functionality to create/add

cicalese:
Replace the standalone 'canasta import' command with integrated database import support. Add --database and --wiki-settings flags to 'canasta create' and 'canasta add' so users can import an existing database dump instead of running install.php. Add --global-settings to 'canasta create' for copying a settings file to config/settings/.

Add 'canasta export' command to export a wiki database (with optional gzip compression) and 'canasta import' (replacing the previous canasta import) command to re-import a database into an existing wiki.

Fix database operations: use docker compose cp instead of bare docker cp, read DB password from .env instead of hardcoding, escape special characters in passwords, use --no-defaults to avoid MySQL config conflicts, and omit mysqldump --databases flag to prevent USE statements in dumps.

Remove dead code: ExportDatabase (replaced by dump command), CopyEnvFile, CopyLocalSettings, CopyDatabase, and SanityChecks (only used by old import).

Related changes:
* CanastaWiki/Canasta-CLI#149
* CanastaWiki/Canasta-Documentation#24


yaronkoren:
This approach makes a lot of sense! There's no point having a two-step process if one step will do.

yaronkoren (approved):

PR #168, opened 2026-02-01
Fix: Don't wrap DB passwords in quotes in .env file

### LINKED ISSUE #145: Error while creating a Canasta wiki using canasta create command
**Command used:**
`sudo canasta create --id canasta-wiki --admin MyName -w CanastaWiki -n localhost -s "Password@123." -v
`
Error: 
`Cannot access the database: :real_connect(): (HY000/1045): Access denied for user 'root'@'canasta-wiki-web-1.canasta-wiki_default' (using password: YES). Check the host, username and password and try again. If using "localhost" as the database host, try using "127.0.0.1" instead (or vice versa).`

### LINKED ISSUE #167: Database password mismatch when password contains quote characters
## Problem

When a generated database password contains a `"` character, there is a mismatch between what MySQL receives and what the CLI uses for database operations.

## Root Cause

In `CreateEnvFile()`, passwords are wrapped in quotes and internal quotes are escaped:
```go
pass := "\"" + strings.ReplaceAll(rootDBpass, "\"", "\\\"") + "\""
```

This causes a mismatch because:
- Docker-compose processes `\"` as an escape sequence → `abc"123`
- The CLI just strips outer quotes but doesn't process escape sequences → `abc\"123`

## Example

For a password `abc"123`:
- Written to .env: `MYSQL_PASSWORD="abc\"123"`
- Docker-compose passes to MySQL: `abc"123` (processes `\"` as literal `"`)
- CLI reads and uses: `abc\"123` (strips quotes but keeps backslash)
- Result: Access denied error

## Impact

This bug occurs randomly depending on whether the generated password contains a `"` character. Since passwords include symbols, `"` is one of the possible characters.

## Fix

1. Don't wrap passwords in quotes when writing to .env
2. Don't escape embedded `"` characters with `\"`

Unquoted values with literal `"` characters work correctly for both docker-compose and the CLI.

## Related

This fix will also resolve #145.

cicalese:
## Summary

Fixes multiple issues with password handling in .env files:

1. **Don't wrap passwords in quotes** - Docker-compose and the CLI handled quoted values differently, causing password mismatches when passwords contained `"` characters

2. **Handle `=` in passwords** - Use `SplitN` to only split on the first `=` when parsing .env files, so passwords containing `=` are read correctly

3. **Use restricted character set for generated passwords** - Only use safe symbols (`@%^-_+.,:`) that won't cause issues in .env files or shell commands

4. **Warn about problematic characters in user-provided passwords** - If a user provides a password via `--envfile` that contains potentially problematic characters, print a warning

## Problem

Randomly occurring "Access denied" errors when the generated password happened to contain characters like `"` or `=`:
- `"` caused docker-compose and CLI to read different values due to escape sequence handling
- `=` caused the CLI to truncate the password at the first `=` in the value

## Test plan

- [x] Create a new installation and verify database operations work
- [x] Verify .env file contains unquoted MYSQL_PASSWORD value
- [x] Verify generated passwords only contain safe characters

Fixes #167
Fixes #145

### DIFF
--- a/internal/canasta/canasta.go
+++ b/internal/canasta/canasta.go
@@ -102,6 +102,14 @@ func CreateEnvFile(customEnvPath, installPath, workingDir, rootDBpass, wikiDBpas
 		// Read custom env file
 		customEnv := GetEnvVariable(customEnvPath)
 
+		// Check for problematic characters in password fields
+		if pw, ok := customEnv["MYSQL_PASSWORD"]; ok {
+			warnIfProblematicPassword("MYSQL_PASSWORD", pw)
+		}
+		if pw, ok := customEnv["WIKI_DB_PASSWORD"]; ok {
+			warnIfProblematicPassword("WIKI_DB_PASSWORD", pw)
+		}
+
 		// Apply each override to .env
 		for key, value := range customEnv {
 			if err := SaveEnvVariable(installPath+"/.env", key, value); err != nil {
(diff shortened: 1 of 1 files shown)


jeffw16 (approved):

yaronkoren:
These look like great improvements!

PR #190, opened 2026-02-05
Improve upgrade output and fix MW_SECRET_KEY handling during create

### LINKED ISSUE #189: Improve upgrade output and fix MW_SECRET_KEY handling during create
### 1. Upgrade output is uninformative
Running `canasta upgrade` gives no indication of what was actually upgraded. The dry-run mode shows detailed file lists, but the normal mode just silently applies changes.

### 2. MW_SECRET_KEY not set in .env during fresh install
When `canasta create` runs the MediaWiki installer, the generated `$wgSecretKey` is left in `config/LocalSettings.php` instead of being extracted to `.env` as `MW_SECRET_KEY`. The new architecture expects `MW_SECRET_KEY` in `.env` and no `config/LocalSettings.php`.

### 3. User-provided MW_SECRET_KEY can be overwritten
If a user provides `MW_SECRET_KEY` via the `-e` env file flag, both the installer path and the database import path overwrite it.

### 4. install.php fails if MW_SECRET_KEY is in the environment
When `MW_SECRET_KEY` is already in `.env` (e.g. from the `-e` flag), the container has it in its environment, causing `CanastaDefaultSettings.php` to think the wiki is already configured, which prevents `install.php` from running.

## Changes

- Show files updated/removed and container images changed during upgrade
- Extract `$wgSecretKey` from installer-generated `LocalSettings.php` to `.env` and delete `LocalSettings.php`
- Use `env -u MW_SECRET_KEY` when running `install.php` so it works regardless of environment state
- Respect user-provided `MW_SECRET_KEY` in both installer and database import paths

cicalese:
## Summary

- **Upgrade output**: Show files updated/removed from upstream, which container images changed, and a clear final summary distinguishing actual upgrades from already up-to-date installations
- **MW_SECRET_KEY**: Extract `$wgSecretKey` from installer-generated `LocalSettings.php` into `.env` and delete `LocalSettings.php` (new architecture); use `env -u MW_SECRET_KEY` so `install.php` runs regardless of environment state; respect user-provided key from `-e` flag

## Test plan

- [x] `canasta create -i test` creates instance with `MW_SECRET_KEY` in `.env` and no `config/LocalSettings.php`
- [x] `canasta create -i test -e custom.env` with `MW_SECRET_KEY` in `custom.env` preserves the user-provided key
- [x] `canasta create -i test -d dump.sql` with database import respects existing `MW_SECRET_KEY`
- [x] `canasta upgrade` on up-to-date instance shows clean output with no spurious messages
- [x] `canasta upgrade` after pushing upstream changes shows files updated and images changed
- [x] `canasta upgrade --dry-run` still shows expected preview output

Fixes #189

### DIFF
--- a/cmd/create/create.go
+++ b/cmd/create/create.go
@@ -186,13 +186,22 @@ func createCanasta(canastaInfo canasta.CanastaVariables, workingDir, path, wikiI
 	// Determine the base image to use
 	var baseImage string
 	if buildFromPath != "" {
-		// Build Canasta (and optionally CanastaBase) from source
-		logging.Print("Building Canasta from local source...\n")
-		builtImage, err := imagebuild.BuildFromSource(buildFromPath)
-		if err != nil {
-			return fmt.Errorf("failed to build from source: %w", err)
+		// Check if local Canasta repo exists for building
+		canastaPath := filepath.Join(buildFromPath, "Canasta")
+		if _, err := os.Stat(canastaPath); err == nil {
+			// Build Canasta (and optionally CanastaBase) from source
+			logging.Print("Building Canasta from local source...\n")
+			builtImage, err := imagebuild.BuildFromSource(buildFromPath)
+			if err != nil {
+				return fmt.Errorf("failed to build from source: %w", err)
+			}
+			baseImage = builtImage
+		} else {
+			// No local Canasta repo, use registry image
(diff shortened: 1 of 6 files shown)


yaronkoren (approved):

yaronkoren:
Oh, never mind, I didn't notice the Copilot "objections".

PR #258, opened 2026-02-16
K8s 3/5: Add kustomization.yaml generation

cicalese:
## Summary

Part of #131. **PR 3 of 5** — merge order: #254 → #256 → #257 → **#258** → #259 → #260

Adds `GenerateKustomization()` which creates a per-installation `kustomization.yaml` with the namespace set to the installation ID. This is analogous to `RewriteCaddy()` — called after config changes.

### Files

- New: `internal/canasta/kustomize.go` — `GenerateKustomization(installPath, namespace, image)`
- New: `internal/canasta/files/kustomization.yaml.tmpl` — embedded template with namespace, resources, configMapGenerator, and conditional `images:` section
- New: `internal/canasta/kustomize_test.go` — tests for generation, image override, and overwrite behavior

### Template features

- `namespace:` set to installation ID
- Resources pointing to `kubernetes/*.yaml`
- `configMapGenerator` for config files and `.env`
- Optional `images:` section for custom container images (used by registry support in PR 5)

### Test plan

- [ ] `go build ./...` passes
- [ ] `go test ./...` passes
- [ ] `golangci-lint run ./...` passes

PR #254, opened 2026-02-16
Refactor and rename restic command to backup

cicalese:
## Summary

Closes #253

Refactors all backup subcommands to use the orchestrator abstraction instead of building Docker commands directly, and renames the top-level command from `restic` to `backup` to be less dependent on a specific backup implementation.

### Key changes

- **`RunRestic` as interface method**: Added to the `Orchestrator` interface, implemented by `ComposeOrchestrator`. This follows the same polymorphic pattern as all other orchestrator operations — no type switches.
- **Eliminated `commandArgs` global**: The old pattern assembled a `[]string` of Docker `run` arguments across the package. Now each subcommand calls `orch.RunRestic()` through a thin wrapper.
- **Simplified subcommands**: Each subcommand file is now a thin wrapper around `runRestic()` instead of manually building Docker argument lists.
- **Renamed command**: `canasta restic` → `canasta backup`
- **Renamed subcommands** for clarity:
  - `take-snapshot` → `create`
  - `view` → `list`
  - `list` → `files`
  - `forget` → `delete`
- Unchanged subcommands: `init`, `restore`, `check`, `diff`, `unlock`, `schedule`

### Files changed

| File | Change |
|------|--------|
| `internal/orchestrators/orchestrator.go` | Add `RunRestic` to `Orchestrator` interface |
| `internal/orchestrators/compose.go` | Implement `RunRestic` on `ComposeOrchestrator` |
| `cmd/backup/backup.go` | New main command (was `cmd/restic/restic.go`) |
| `cmd/backup/*.go` (9 subcommands) | Renamed package, updated command names and text |
| `cmd/root/root.go` | Update import path and registration |
| `docs/guide/backup.md` | Update command references |
| `docs/guide/best-practices.md` | Update command references |
| `mkdocs.yml` | Update CLI reference nav |
| `internal/orchestrators/orchestrator_test.go` | Add `RunRestic` to mock |
| `cmd/maintenanceUpdate/extension_test.go` | Add `RunRestic` to mock |

## Test plan

- [ ] `go build ./...` passes
- [ ] `go test ./...` passes
- [ ] `golangci-lint run ./...` passes
- [ ] Manual test: `canasta backup create`, `canasta backup list`, `canasta backup restore` work with Docker Compose

Issue #131, opened 2025-11-13
Add support for Kubernetes


yaronkoren:
It would be great for the CLI to support installations that use Canasta-Kubernetes instead of Canasta-DockerCompose. This may not require any changes to the CLI syntax - rather, it may just involve making sure that every existing CLI command also works when the Kubernetes option is used.

tosfos:
## Analysis of Kubernetes Support Implementation

I've analyzed the current Canasta-CLI codebase and the existing Canasta-Kubernetes repository. Here's my assessment of what would be needed to add Kubernetes support:

### Current Architecture
The CLI already has a solid orchestrator abstraction layer in `internal/orchestrators/orchestrators.go` that supports different container orchestration systems. Currently, only "compose" (Docker Compose) is implemented.

### Key Findings

**✅ Good News - Architecture is Ready:**
- The orchestrator interface already exists
- Repository cloning logic in `CloneStackRepo()` can handle different orchestrator types
- Configuration system supports multiple orchestrators
- CLI already accepts `-o/--orchestrator` flag (defaults to "compose")

**✅ Canasta-Kubernetes Repository Exists:**
- Complete Kubernetes manifests are available at https://github.com/CanastaWiki/Canasta-Kubernetes
- Includes deployments for web, db, varnish, elasticsearch, caddy
- Supports both manual `kubectl` deployment and ArgoCD GitOps
- Uses Kustomize for configuration management

### Required Changes

**1. Update `GetRepoLink()` function:**
```go
func GetRepoLink(orchestrator string) string {
    switch orchestrator {
    case "docker-compose", "compose":
        return "https://github.com/CanastaWiki/Canasta-DockerCompose.git"
    case "kubernetes":
        return "https://github.com/CanastaWiki/Canasta-Kubernetes.git"
    default:
        logging.Fatal(fmt.Errorf("orchestrator: %s is not available", orchestrator))
    }
    return ""
}
```

**2. Add Kubernetes implementations in orchestrators.go:**
- `Start()` - Use `kubectl apply -k .` or `kubectl apply -f argocd-app.yaml`
- `Stop()` - Use `kubectl delete` commands
- `Exec()` - Use `kubectl exec`
- `CheckDependencies()` - Check for `kubectl` instead of `docker compose`
- Database operations would need Kubernetes equivalents

**3. Update config validation:**
- Modify `AddOrchestrator()` to accept "kubernetes" as valid orchestrator
[...]

PR #259, opened 2026-02-16
K8s 4/5: Add Kubernetes restic backup support

cicalese:
## Summary

Part of #131. **PR 4 of 5** — merge order: #254 → #256 → #257 → #258 → **#259** → #260

Adds Kubernetes-specific restic backup support and guards cron scheduling for K8s installations.

### Changes

- **`internal/orchestrators/kubernetes.go`**: `RunRestic` implementation — runs restic as an ephemeral pod via `kubectl run` with optional PVC volume mounts via `--overrides`
- **`cmd/restic/schedule.go`**: Guard for K8s — returns error directing users to edit the backup CronJob in kustomization.yaml instead of using host cron

### How K8s restic works

```
kubectl run restic-cli --rm -i --restart=Never -n {namespace} \
  --image=restic/restic --env-from=configmap/canasta-env \
  [--overrides=<pod spec with volume mounts>] \
  -- <restic args>
```

### Test plan

- [ ] `go build ./...` passes
- [ ] `go test ./...` passes
- [ ] `golangci-lint run ./...` passes
- [ ] Compose restic commands unchanged

### DIFF
--- a/cmd/backup/schedule.go
+++ b/cmd/backup/schedule.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/CanastaWiki/Canasta-CLI/internal/logging"
+	"github.com/CanastaWiki/Canasta-CLI/internal/orchestrators"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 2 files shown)


cicalese:
Superseded by #315. K8s backup support rebased onto the embedding stack.

PR #275, opened 2026-02-17
Fix create command destroying existing installation on duplicate ID

### LINKED ISSUE #274: canasta create destroys existing installation when ID already exists
## Bug

When running `canasta create -i <id>` with an ID that already exists, the command correctly detects the duplicate at the start of `createCanasta()` and returns an error. However, the `RunE` handler in the caller treats **all** errors from `createCanasta` as partial-installation failures and runs `DeleteConfigAndContainers` to clean up — which destroys the existing, working installation.

## Steps to reproduce

```bash
canasta create -i wikifarm -w one -a admin
# ... installation succeeds

canasta create -i wikifarm -w one -a admin
# Output: "Canasta installation with the ID already exist!"
# Then: "Removing containers" / "Deleting config files" / "Deleted all containers and config files"
```

The existing `wikifarm` installation is now gone.

## Root cause

In `cmd/create/create.go`, the error handler at lines 143-148 unconditionally calls `canasta.DeleteConfigAndContainers()` for any error returned by `createCanasta()`, including the early "ID already exists" check. The cleanup is intended for partial installations that failed mid-setup, not for pre-validation errors.

## Expected behavior

When the ID already exists, the command should print an error and exit without modifying the existing installation.

cicalese:
## Summary

- Moves the duplicate instance ID check from inside `createCanasta()` to the `RunE` validation block, so it returns an error immediately without reaching the cleanup handler
- Previously, the "ID already exists" error was treated as a partial-installation failure, causing `DeleteConfigAndContainers` to destroy the existing working installation

Fixes #274

## Test plan

- [ ] `canasta create -i existing-id ...` with an existing ID prints error and exits without modifying the existing installation
- [ ] `canasta create -i new-id ...` still works normally
- [ ] A failed installation mid-setup (e.g. bad domain) still cleans up as before
- [ ] `go test ./...` passes

### DIFF
--- a/cmd/create/create.go
+++ b/cmd/create/create.go
@@ -79,6 +79,11 @@ instead of running the installer, or enable development mode with Xdebug.`,
 				return fmt.Errorf("Canasta instance ID should not contain spaces or non-ASCII characters, only alphanumeric characters are allowed")
 			}
 
+			// Check for duplicate ID before doing any work
+			if _, err := config.GetDetails(canastaInfo.Id); err == nil {
+				return fmt.Errorf("Canasta installation with ID '%s' already exists", canastaInfo.Id)
+			}
+
 			// Validate --dev-tag and --build-from are mutually exclusive
 			if devTag != "latest" && buildFromPath != "" {
 				return fmt.Errorf("--dev-tag and --build-from are mutually exclusive")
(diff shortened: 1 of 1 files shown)


yaronkoren (approved):
Ooh, that check was not too helpful in its previous location!

cicalese:
It was the opposite of helpful - lol. Yes, I did blow away a running installation by accidentally running `canasta create` a second time.

PR #319, opened 2026-02-21
Run maintenance update after enabling extension(s) or skin(s)

### LINKED ISSUE #67: SMW Error is displayed when SemanticMediawiki is enabled during setup Wizard
Reproducible on Canasta's latest image tag, which is [1.43.5-20251202-cd387ac](https://github.com/CanastaWiki/Canasta/pkgs/container/canasta/597078880?tag=1.43.5-20251202-cd387ac) and Canasta-DockerCompose repo of commit [ffb4b3d7c4ee3eb8399acc2415080ee79a293ee8](ffb4b3d7c4ee3eb8399acc2415080ee79a293ee8) at the moment of the bug report

Steps to reproduce:

* Create a new Canasta wiki (i.e., using https://github.com/CanastaWiki/Canasta-DockerCompose)
* During the setup wizard, check the SemanticMediawiki extension checkbox
* Complete the setup & store the generated `LocalSettings.php` into the  `config/` directory as the guide says [[1]](https://canasta.wiki/manual-download/) or other directory that's mounted into `/mediawiki/config`
* Visit the wiki URL

Expected result:

* The wiki homepage is shown

Actual result:

* `Semantic MediaWiki was installed and enabled, but is missing an appropriate upgrade key.` error is displayed

Image

* Restarting the stack does not help
* Running `update.php` in the container helps

Likely related to:
* https://github.com/CanastaWiki/Canasta-DockerCompose/issues/66

naresh-kumar-babu:
Fixes https://github.com/CanastaWiki/Canasta-DockerCompose/issues/67

### DIFF
--- a/cmd/extension/enable.go
+++ b/cmd/extension/enable.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	maintenance "github.com/CanastaWiki/Canasta-CLI/cmd/maintenanceUpdate"
 	"github.com/CanastaWiki/Canasta-CLI/internal/extensionsskins"
 	"github.com/spf13/cobra"
 )
@@ -24,9 +25,10 @@ extension for a specific wiki only.`,
 
   # Enable an extension for a specific wiki
   canasta extension enable VisualEditor -i myinstance -w docs`,
-		Args:  cobra.ExactArgs(1),
+		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			extensions := strings.Split(args[0], ",")
+			mainatenanceUpdateRequired := false
 			for _, extension := range extensions {
 				extensionName, err := extensionsskins.CheckInstalled(extension, instance, orch, constants)
 				if err != nil {
@@ -35,6 +37,15 @@ extension for a specific wiki only.`,
 				}
 				if err := extensionsskins.Enable(extensionName, wiki, instance, orch, constants); err != nil {
 					return err
+				}
+				if extensionName == "SemanticMediaWiki" {
+					mainatenanceUpdateRequired = true
+				}
+			}
+			if mainatenanceUpdateRequired {
+				// Run maintenance update
+				if err := maintenance.RunMaintenanceUpdate(instance, wiki); err != nil {
+					return fmt.Errorf("maintenance update failed after enabling extension(s): %v", err)
 				}
 			}
 			return nil
--- a/cmd/extension/enable.go
+++ b/cmd/extension/enable.go
@@ -4,8 +4,8 @@ import (
 	"fmt"
 	"strings"
 
-	maintenance "github.com/CanastaWiki/Canasta-CLI/cmd/maintenanceUpdate"
 	"github.com/CanastaWiki/Canasta-CLI/internal/extensionsskins"
+	internalmaintenance "github.com/CanastaWiki/Canasta-CLI/internal/maintenance"
 	"github.com/spf13/cobra"
 )
 
@@ -25,7 +25,7 @@ extension for a specific wiki only.`,
 
   # Enable an extension for a specific wiki
   canasta extension enable VisualEditor -i myinstance -w docs`,
-		Args: cobra.ExactArgs(1),
+		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			extensions := strings.Split(args[0], ",")
 			for _, extension := range extensions {
@@ -38,8 +38,14 @@ extension for a specific wiki only.`,
 					return err
 				}
 			}
-			if err := maintenance.RunMaintenanceUpdate(instance, wiki); err != nil {
-				return fmt.Errorf("maintenance update failed after enabling extension(s): %v", err)
+			var maintErr error
+			if wiki != "" {
+				maintErr = internalmaintenance.RunMaintenanceUpdate(instance, wiki, internalmaintenance.Options{})
+			} else {
+				maintErr = internalmaintenance.RunUpdateAllWikis(instance, internalmaintenance.Options{})
+			}
+			if maintErr != nil {
+				return fmt.Errorf("maintenance update failed after enabling extension(s): %v", maintErr)
 			}
 			return nil
 		},
--- a/cmd/extension/enable.go
+++ b/cmd/extension/enable.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/CanastaWiki/Canasta-CLI/cmd/maintenanceUpdate"
+	maintenance "github.com/CanastaWiki/Canasta-CLI/cmd/maintenanceUpdate"
 	"github.com/CanastaWiki/Canasta-CLI/internal/extensionsskins"
 	"github.com/spf13/cobra"
 )
@@ -38,7 +38,7 @@ extension for a specific wiki only.`,
 					return err
 				}
 			}
-			if err := maintenanceUpdate.RunMaintenanceUpdate(instance, wiki); err != nil {
+			if err := maintenance.RunMaintenanceUpdate(instance, wiki); err != nil {
 				return fmt.Errorf("maintenance update failed after enabling extension(s): %v", err)
 			}
 			return nil
--- a/cmd/extension/enable.go
+++ b/cmd/extension/enable.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


cmd/extension/enable.go
```diff
@@ -36,6 +38,15 @@ extension for a specific wiki only.`,
 				if err := extensionsskins.Enable(extensionName, wiki, instance, orch, constants); err != nil {
 					return err
 				}
+				if extensionName == "SemanticMediaWiki" {
```
jeffw16: SMW isn’t the only extension that requires update.php to be run. There should be a more general check for whether the extension requiresupdate.php to be run to account for database changes.

jeffw16 (changes requested):

cicalese (commented):
Review from Claude:

This PR solves a real problem — SMW requires `update.php` to create its database tables, and `canasta extension enable` currently doesn't run it. However, there are a few concerns:

**1. Hardcoded extension name is brittle**

Checking `extensionName == "SemanticMediaWiki"` means any other extension that needs schema changes (Cargo, CirrusSearch, etc.) would need the same special case. Consider running `update.php` after enabling *any* extension — it's a safe no-op when no schema changes are needed, and it future-proofs the command.

**2. Typo**

`mainatenanceUpdateRequired` has an extra `a`.

**3. Cross-package coupling**

Importing `cmd/maintenanceUpdate` from `cmd/extension` creates a dependency between command packages. If `RunMaintenanceUpdate` is needed from multiple places, the core logic should live in an `internal/` package rather than being exported from a cmd package.

**4. `RunMaintenanceUpdate` creates its own orchestrator**

`RunMaintenanceUpdate` calls `orchestrators.New(instance.Orchestrator)` internally. This doesn't restore K8s-specific state (e.g., `ManagedCluster`). Should use `orchestrators.NewFromInstance(instance)` once the K8s PR chain merges.

**5. Scope question**

The linked issue (CanastaWiki/Canasta-DockerCompose#67) is about the setup wizard in raw Docker Compose, not the CLI. The CLI path doesn't restart after `extension enable`, so the fix is needed here — but worth noting that the original bug report is a different code path.

**Suggested approach:**

Instead of special-casing SMW, run `update.php` unconditionally after enabling extensions. Move the maintenance update logic to an internal package, and keep cmd packages from importing each other.

cmd/extension/enable.go
naresh-kumar-babu: I changed the logic to run maintenance update for all extensions - just to be safe.

naresh-kumar-babu:
> Review from Claude:
> 
> This PR solves a real problem — SMW requires `update.php` to create its database tables, and `canasta extension enable` currently doesn't run it. However, there are a few concerns:
> 
> **1. Hardcoded extension name is brittle**
> 
> Checking `extensionName == "SemanticMediaWiki"` means any other extension that needs schema changes (Cargo, CirrusSearch, etc.) would need the same special case. Consider running `update.php` after enabling _any_ extension — it's a safe no-op when no schema changes are needed, and it future-proofs the command.
> 
> **2. Typo**
> 
> `mainatenanceUpdateRequired` has an extra `a`.
> 
> **3. Cross-package coupling**
> 
> Importing `cmd/maintenanceUpdate` from `cmd/extension` creates a dependency between command packages. If `RunMaintenanceUpdate` is needed from multiple places, the core logic should live in an `internal/` package rather than being exported from a cmd package.
> 
> **4. `RunMaintenanceUpdate` creates its own orchestrator**
> 
> `RunMaintenanceUpdate` calls `orchestrators.New(instance.Orchestrator)` internally. This doesn't restore K8s-specific state (e.g., `ManagedCluster`). Should use `orchestrators.NewFromInstance(instance)` once the K8s PR chain merges.
> 
> **5. Scope question**
> 
> The linked issue ([CanastaWiki/Canasta-DockerCompose#67](https://github.com/CanastaWiki/Canasta-DockerCompose/issues/67)) is about the setup wizard in raw Docker Compose, not the CLI. The CLI path doesn't restart after `extension enable`, so the fix is needed here — but worth noting that the original bug report is a different code path.
> 
> **Suggested approach:**
> 
> Instead of special-casing SMW, run `update.php` unconditionally after enabling extensions. Move the maintenance update logic to an internal package, and keep cmd packages from importing each other.

1. Fixed
2. Fixed
3. It is only used in one another place apart from the place of definition. I don't think it is a heavy reuse candidate.
4. As this point is dependent on a future change, I think addressing it post that merge in a separate PR would be better.
5. Since we don't plan to use Canasta-DockerCompose directly anymore, I think making this fix in CLI is the right way to go.

cicalese:
Thank you for the update. Here is some additional feedback:

1. When no -w flag is given, Enable() installs the extension globally (all wikis), but RunMaintenanceUpdate(instance, wiki) is called with an empty wiki, which runs update.php without --wiki. This runs against whatever PHP defaults to, not all wikis. It should get the wiki list and run update.php on each one, like the --all path in the maintenance-update command does.
2. The PR only adds maintenance update to extension enable. Skins can also require database changes (though less common), and extension disable / skin disable could also benefit from a maintenance update to clean up. At minimum, skin enable has the same need.
3. skipJobs is a package-level var in maintenanceUpdate that's only set via the --skip-jobs flag on the maintenance-update command. When RunMaintenanceUpdate is called from extension enable, skipJobs will always be its zero value (false), so runJobs.php will always run. This may be fine (running jobs after enabling is reasonable), but the caller has no way to control it.

jeffw16:
@naresh-kumar-babu:

The code review said:

> Importing cmd/maintenanceUpdate from cmd/extension creates a dependency between command packages. If RunMaintenanceUpdate is needed from multiple places, the core logic should live in an internal/ package rather than being exported from a cmd package.

Your response:

> It is only used in one another place apart from the place of definition. I don't think it is a heavy reuse candidate.

Here's my concerns.

This isn’t about reuse count — it’s about architectural boundaries. `cmd` is the composition root in DI terms. Importing from it turns the composition layer into a dependency, which is backwards. Let’s move the logic into `internal/...` and keep `cmd` as wiring only.

Elaboration:

Even if it’s only called from one other place, importing from `cmd/...` reverses our dependency direction: `cmd` should be a thin entrypoint layer, not a library. This coupling makes refactors and testing harder and encourages `cmd` to accumulate shared logic.

From a dependency injection perspective, `cmd` is where we compose dependencies (flags, config, wiring), not where we define reusable behavior. Core logic should depend on abstractions and receive its dependencies via parameters. If other packages import from `cmd`, we’re effectively making the composition layer part of our application API, which breaks DI boundaries and makes mocking/testing more awkward.

The core behavior should live in `internal/...` (or another non-`cmd` package), and `cmd/maintenanceUpdate` should just wire dependencies and call into that logic.

cmd/extension/enable.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	maintenance "github.com/CanastaWiki/Canasta-CLI/cmd/maintenanceUpdate"
```
jeffw16: cmd should not import from other cmd files, as this violates dependency injection principles and makes things hard to test/debug

jeffw16 (changes requested):

cmd/extension/enable.go
naresh-kumar-babu: Moved the logic to `internal/`

naresh-kumar-babu:
> Thank you for the update. Here is some additional feedback:
> 
> 1. When no -w flag is given, Enable() installs the extension globally (all wikis), but RunMaintenanceUpdate(instance, wiki) is called with an empty wiki, which runs update.php without --wiki. This runs against whatever PHP defaults to, not all wikis. It should get the wiki list and run update.php on each one, like the --all path in the maintenance-update command does.
> 2. The PR only adds maintenance update to extension enable. Skins can also require database changes (though less common), and extension disable / skin disable could also benefit from a maintenance update to clean up. At minimum, skin enable has the same need.
> 3. skipJobs is a package-level var in maintenanceUpdate that's only set via the --skip-jobs flag on the maintenance-update command. When RunMaintenanceUpdate is called from extension enable, skipJobs will always be its zero value (false), so runJobs.php will always run. This may be fine (running jobs after enabling is reasonable), but the caller has no way to control it.

I’ve made the necessary changes. Please review the latest changes.

yaronkoren:
I believe this PR now has to be rewrriten, or replaced entirely, unfortunately, after some recent refactoring of the code, most notably in #420.

cicalese:
Thank you for your work on this @naresh-kumar-babu.

Superseded by #464. The codebase has changed significantly since this PR was opened (enable/disable consolidated into internal/extensionsskins, maintenance refactored). A fresh implementation based on this start will be cleaner than rebasing.

cicalese:
Replaced by #465.

PR #348, opened 2026-02-22
Add --cleanup flag to canasta list

### LINKED ISSUE #330: Add --cleanup flag to canasta list to remove stale config entries
If an installation directory is deleted outside the CLI (manual cleanup, failed install that partially registered, etc.), the entry remains in the config registry (`conf.json`) with no way to remove it. Commands that look up the installation will fail with confusing errors.

`canasta list --cleanup` should:
- Check whether each registered installation's directory still exists
- Remove entries whose directories are missing
- Print a message for each removal, e.g. "Removed stale entry for instance 'X' (directory not found)"
- List the remaining valid instances afterward

Without `--cleanup`, `canasta list` could still note missing directories in its output (e.g., an extra column or marker) so the user is aware of the problem.

cicalese:
## Summary
- Adds `--cleanup` flag to `canasta list` that removes stale config entries whose installation directories no longer exist
- For Kubernetes installations with managed kind clusters, `--cleanup` also deletes the orphaned kind cluster
- Without `--cleanup`, stale entries are marked with `[not found]` in the Installation Path column
- When no installations exist, prints "No instances found." instead of an empty table header

Closes #330

## Test plan
- [x] Add a fake entry to conf.json, run `canasta list` — should show `[not found]` marker
- [x] Run `canasta list --cleanup` — should remove the stale entry and print removal message
- [x] For a stale k8s entry with a kind cluster, verify the cluster is also deleted
- [x] Run `canasta list` again — stale entry should be gone
- [x] Run `canasta list` with no installations — should print "No instances found."

### DIFF
--- a/cmd/list/list.go
+++ b/cmd/list/list.go
@@ -1,30 +1,59 @@
 package list
 
 import (
+	"fmt"
+	"os"
+
 	"github.com/spf13/cobra"
 
 	"github.com/CanastaWiki/Canasta-CLI/internal/config"
+	"github.com/CanastaWiki/Canasta-CLI/internal/orchestrators"
 )
 
-var instance config.Installation
+var (
+	instance config.Installation
+	cleanup  bool
+)
 
 func NewCmdCreate() *cobra.Command {
 	var listCmd = &cobra.Command{
 		Use:   "list",
 		Short: "List all Canasta installations",
 		Long: `List all registered Canasta installations. Displays each installation's
-ID, path, and orchestrator as recorded in the Canasta configuration file.`,
-		Example: `  canasta list`,
+ID, path, and orchestrator as recorded in the Canasta configuration file.
+
+Use --cleanup to remove stale entries whose installation directories no longer exist.`,
+		Example: `  canasta list
+  canasta list --cleanup`,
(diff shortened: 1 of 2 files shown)


yaronkoren:
The overall concept is great - but this does make sense as a feature of "canasta list"? I've never heard of a list command that makes changes. I don't know if there's a better place to put it, though... maybe a "canasta cleanup" command?

cicalese:
I had considered a separate cleanup command, but it seemed like overkill. I need this when I'm trying to start an instance and it says the name is already in use or the ports are already in use. So, this patch does two things: it lets you know in the listing that although the instance is listed in the config, it no longer actually exists; AND it lets you clean up, after which it does another list of the remaining instances. It feels correct that, if it is already figuring out which instances are orphaned for the listing, it could optionally clean them up. It is useful to the sysadmin this way, and adding another command would overcomplicate this and make it less useful, since you'd need to call another command. And, there'd be duplication of code, since it most of what it would be doing would be essentially the same.

yaronkoren:
Alright, sounds good.

yaronkoren (approved):

PR #358, opened 2026-02-24
Implement backup and restore for Kubernetes installations

### LINKED ISSUE #324: Implement backup and restore for Kubernetes installations
All `canasta backup` subcommands (init, create, restore, list, delete, unlock, files, check, diff, schedule) currently return "backup is not yet supported for Kubernetes installations".

This is the main functional gap between the Compose and Kubernetes orchestrators.

## Scope decisions

- **Remote repositories only.** The K8s orchestrator will support remote restic repositories (S3, SFTP, etc.) but not local filesystem paths. Local repos require exposing host paths into Jobs via hostPath volumes, which is fragile and not portable. The Compose orchestrator will continue to support both local and remote repos.

- **CronJob for scheduling.** Instead of host crontab (used by Compose), the K8s orchestrator will use a native Kubernetes CronJob resource for `canasta backup schedule`. This is the idiomatic K8s approach and avoids host-level side effects. See also #278.

## Implementation approach

The Compose backup implementation is the reference. It stages data into a Docker volume, runs restic against it, and cleans up. The K8s equivalent:

1. **Database export:** `kubectl exec` into the db pod to run `mysqldump` (similar to Compose's `docker compose exec`)
2. **Staging:** Create a temporary Job that mounts source PVCs (config-data hostPath, mysql-data PVC, sitemap-data PVC) read-only and a staging PVC read-write, copies the backup payload into the staging PVC
3. **Restic:** Create a Job that mounts the staging PVC and runs restic with remote repo credentials from environment variables
4. **Restore:** Reverse of staging — restic restores to staging PVC, then a Job copies files back to source PVCs
5. **Cleanup:** Delete staging PVC and completed Jobs
6. **Schedule:** Create a CronJob resource that runs the backup Job on the user's requested schedule

## What needs to be backed up

- Database dumps: `db_{wikiid}.sql` per wiki
- Configuration: `config/` directory
- Extensions: `extensions/` directory
- Images: `images/` directory
- Skins: `skins/` directory
- Public assets: `public_assets/` directory
- Special files: `.env`, `my.cnf`

cicalese:
## Summary

- Extract shared backup helpers (volume staging, restic execution, restore) from `compose.go` into `backup.go` so both orchestrators can reuse them
- Implement `RunBackup` for Kubernetes: syncs cluster data (extensions, skins, images, DB dumps) to the CLI host via `kubectl cp`, stages into a Docker volume, and runs restic
- Implement `RestoreFromBackupVolume` for Kubernetes: restores from Docker volume to CLI host, then syncs extensions/skins/images back to the cluster
- Reject local backup repositories for K8s with a clear error (remote backends only)

Closes #324

## Test plan

- [x] `go build ./...` compiles cleanly
- [x] `go test ./...` passes (including new `TestBackupVolumeName` and `TestRunBackupRejectsLocalRepo`)
- [x] Compose backup/restore regression: `canasta backup create` / `canasta backup restore` still work on Compose installations
- [x] K8s backup create: create a kind installation, run `canasta backup init` and `canasta backup create` with a remote repo
- [x] K8s backup restore: run `canasta backup restore` and verify config restored on host, extensions/skins/images restored in cluster, databases restored
- [x] K8s local repo rejection: verify `canasta backup init` with `RESTIC_REPOSITORY=/local/path` returns clear error
- [x] K8s backup list/check/delete/files/diff work against the remote repo

### DIFF
--- a/internal/orchestrators/backup.go
+++ b/internal/orchestrators/backup.go
@@ -1,0 +1,110 @@
+package orchestrators
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/CanastaWiki/Canasta-CLI/internal/execute"
+)
+
+// backupVolumeName returns the persistent Docker volume name for an installation.
+func backupVolumeName(installPath string) string {
+	return "canasta-backup-" + filepath.Base(installPath)
+}
+
+// isLocalRepo returns true if the repository URL is a local filesystem path
+// rather than a remote backend (s3:, sftp:, rest:, gs:, azure:, b2:, rclone:).
+func isLocalRepo(repoURL string) bool {
+	return strings.HasPrefix(repoURL, "/")
+}
+
+// repoFromArgs extracts the repository URL following the -r flag in args.
+func repoFromArgs(args []string) string {
+	for i, arg := range args {
+		if arg == "-r" && i+1 < len(args) {
+			return args[i+1]
+		}
+	}
+	return ""
+}
+
+// stageToVolume copies host directories into the backup volume via an alpine container.
+func stageToVolume(volName string, volumes map[string]string) error {
+	cmdArgs := []string{"docker", "run", "--rm",
+		"-v", volName + ":/currentsnapshot",
+	}
+
+	var copyParts []string
+	i := 0
+	for hostPath, containerPath := range volumes {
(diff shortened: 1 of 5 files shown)


yaronkoren (approved):
Awesome! What a huge improvement to Kubernetes handling.

PR #368, opened 2026-02-24
Add healthcheck for Varnish container

### LINKED ISSUE #367: Add healthcheck for Varnish container
The Varnish service in `docker-compose.yml` has `restart: unless-stopped` but no healthcheck, unlike the db and elasticsearch services which do have healthchecks. This means if Varnish hangs without the container fully crashing (e.g. after an OOM), the restart policy won't trigger.

Adding a healthcheck would improve reliability by detecting hung processes and triggering a restart.

Migrated from CanastaWiki/Canasta-DockerCompose#59.

cicalese:
## Summary

- Adds a healthcheck to the Varnish service in `docker-compose.yml` using `varnishadm ping`
- Detects hung Varnish processes that `restart: unless-stopped` alone would miss (e.g. after OOM)
- Matches the pattern used by the db and elasticsearch services which already have healthchecks

## Test plan

- [x] `go build ./...` and `go test ./...` pass
- [x] `canasta create` — Varnish container starts and reports healthy
- [x] Existing installations pick up the healthcheck after `canasta upgrade`

Closes #367

### DIFF
--- a/internal/orchestrators/compose/files/docker-compose.yml
+++ b/internal/orchestrators/compose/files/docker-compose.yml
@@ -99,6 +99,12 @@ services:
     depends_on:
       web:
         condition: service_healthy
+    healthcheck:
+      test: ["CMD", "varnishadm", "ping"]
+      interval: 10s
+      timeout: 5s
+      retries: 3
+      start_period: 10s
     volumes:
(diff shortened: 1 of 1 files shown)


yaronkoren (approved):
Great, the last of the Canasta-DockerCompose issues is fixed!

PR #375, opened 2026-02-25
Switch database from MySQL 8.0 to MariaDB 11.4

### LINKED ISSUE #354: Switch to using MariaDB
This is a move of the following issue, now that the Canasta-DockerCompose is planned to be archived:

https://github.com/CanastaWiki/Canasta-DockerCompose/issues/1

From the original request/proposal by @jeffw16:

> The Wikimedia Foundation uses MariaDB these days rather than MySQL. I think we should stick to their setup as close as possible. Maybe we could switch to using the MariaDB Docker image rather than MySQL.

cicalese:
## Summary

- Switches the database container from MySQL 8.0 to MariaDB 11.4 LTS (supported until May 2029), aligning with the Wikimedia Foundation stack
- Updates Docker Compose and Kubernetes manifests: image, entrypoint (`mysqld` → `mariadbd`), removes MySQL-specific `--default-authentication-plugin` flag
- Sets `MARIADB_AUTO_UPGRADE=1` so existing MySQL data volumes are automatically upgraded on first start. This is safe to leave permanently — MariaDB only runs the upgrade when needed and skips it on subsequent starts
- Updates `my.cnf` template: removes MySQL 8.0-specific `skip-binary-as-hex`, changes `[mysql]` section to `[client]`
- Updates all CLI commands that exec into the db container to use MariaDB-native binaries (`mariadb`, `mariadb-dump`, `mariadb-admin`) since MariaDB 11.4 does not provide `mysql*` compatibility symlinks
- Updates backup commands (which exec into the web container) to use `mariadb`/`mariadb-dump` for consistency
- Renames observability index pattern from `mysql-logs-*` to `mariadb-logs-*` (Logstash pipeline, Fluent Bit config, OpenSearch init scripts, K8s PVCs)
- Adds MySQL 8.0 → MariaDB dump-and-restore migration to `canasta upgrade` for Docker Compose installations: detects MySQL 8.0 binary data, dumps user databases via a temporary mysql:8.0 container, clears the data volume, lets MariaDB initialize fresh, then imports the dump
- Kubernetes installations display a warning directing users to migrate manually. Automatic migration for Kubernetes is not implemented because it is unlikely that many existing Canasta Kubernetes installations created with the CLI exist to upgrade.
- Updates documentation to reference MariaDB and its native CLI commands

**Unchanged (backwards compatibility):**
- `MYSQL_PASSWORD` and other `MYSQL_*` env vars (MariaDB's Docker image requires these names)
- Docker Compose volume names `mysql-data-volume` and `mysql-logs` (renaming would orphan existing data)
- Port 3306 and log paths (`/var/log/mysql/`)

Closes #354

## Test plan

- [x] `go test ./...` — all tests pass
- [x] `make build` — binary builds successfully
- [x] Create a fresh installation: `canasta create -i test-mariadb -w test -n localhost:8443 -a admin`
- [x] Verify `docker compose exec db mariadbd --version` shows MariaDB 11.4
- [x] Verify all containers healthy and wiki responds
- [x] Test `canasta export` (uses `mariadb-dump` in db container)
- [x] Test `canasta backup create` (uses `mariadb-dump` in web container)
- [x] Test `canasta backup restore` (uses `mariadb` in web container)
- [x] Test `canasta upgrade` on an existing MySQL 8.0 installation — dump-and-restore migration runs successfully


cicalese:
**MySQL 8.0 → MariaDB migration: binary incompatibility**

Switching the db container image from `mysql:8.0` to `mariadb:11.4` works for new installations, but **breaks existing installations** because MySQL 8.0's data files are not binary-compatible with MariaDB.

MySQL 8.0 moved its data dictionary into InnoDB (`mysql.ibd`), a format MariaDB never adopted. MariaDB's `MARIADB_AUTO_UPGRADE` feature only handles upgrades between MariaDB versions — it can't convert MySQL 8.0's data format, and the MariaDB server fails to start against the existing data volume.

The binary compatibility between MySQL and MariaDB only applies to MySQL 5.x. MySQL 8.0 introduced breaking changes.

**Options:**
1. Add a dump-and-restore migration step to `canasta upgrade` — detect MySQL 8.0 data, dump using the old MySQL image, start fresh MariaDB, restore
2. Stay on MySQL (abandon the switch)

New installations are unaffected since MariaDB creates its own data from scratch.

cicalese:
Note that this incompatibility does **not** affect importing MySQL databases via `canasta create --import` or `canasta import`. Those work by running SQL dumps through the MariaDB client, which is a logical import — SQL dumps are compatible across MySQL and MariaDB regardless of version. The incompatibility is specifically with the binary data directory files on disk (`/var/lib/mysql`), i.e. existing installations that are being upgraded in place.

yaronkoren:
Oh, I didn't even think about the possibility that MariaDB couldn't read a MySQL database. But it's not totally suprising.

cicalese:
The MySQL 8.0 → MariaDB migration binary incompatibility has been resolved. The PR now includes a dump-and-restore migration step in `canasta upgrade` for Docker Compose installations: it detects MySQL 8.0 data (`mysql.ibd`), dumps all user databases via a temporary `mysql:8.0` container, clears the data volume, lets MariaDB initialize fresh, and imports the dump. Tested successfully on an existing MySQL 8.0 installation.

Kubernetes installations show a warning directing users to migrate manually, since it's unlikely that many existing Canasta K8s installations created with the CLI exist to upgrade.

yaronkoren (approved):
Incredible! Score for being slightly more open source.

Issue #390, opened 2026-02-26
Make MW_SITEMAP_PAUSE_DAYS configurable via canasta config set


cicalese:
`MW_SITEMAP_PAUSE_DAYS` controls how often the background sitemap generator refreshes sitemaps (default: 1 day). It is currently set only as a Dockerfile `ENV` default in CanastaBase. It is not in the `.env` template or in the `docker-compose.yml` environment passthrough, so it cannot be changed via `canasta config set`.

The sitemap documentation (`docs/guide/sitemaps.md`) already references it as if it belongs in `.env`, but it isn't actually there.

To make it configurable:
1. Add `MW_SITEMAP_PAUSE_DAYS=1` to `internal/canasta/installation-template/.env`
2. Add `- MW_SITEMAP_PAUSE_DAYS=${MW_SITEMAP_PAUSE_DAYS:-1}` to the `web` service environment in `internal/orchestrators/compose/files/docker-compose.yml`
3. Add it to the `canasta config` help text in `cmd/config/config.go`

Ref: https://github.com/CanastaWiki/CanastaBase/pull/99#issuecomment-3952941425

PR #392, opened 2026-02-26
Clarify sitemap docs: skip-redirects, namespace filtering, refresh interval

### LINKED ISSUE #391: Clarify sitemap docs: skip-redirects, namespace filtering, refresh interval
The sitemap documentation should clarify three points raised in https://github.com/CanastaWiki/CanastaBase/pull/99#issuecomment-3952941425:

1. Redirect pages are always excluded from sitemaps (`--skip-redirects` is hardcoded)
2. `$wgSitemapNamespaces` can be used to filter which namespaces are included
3. The `MW_SITEMAP_PAUSE_DAYS` description should not imply the variable is already in `.env` (making it configurable via `canasta config set` is tracked in #390)

cicalese:
## Summary

- Document that redirect pages are always excluded from sitemaps (`--skip-redirects` is hardcoded)
- Document that `$wgSitemapNamespaces` can be used to filter which namespaces are included
- Fix the `MW_SITEMAP_PAUSE_DAYS` description to not imply it's already in `.env` (making it configurable is tracked in #390)

Closes #391

Ref: https://github.com/CanastaWiki/CanastaBase/pull/99#issuecomment-3952941425

## Test plan

- [x] Verify documentation renders correctly

### DIFF
--- a/docs/guide/sitemaps.md
+++ b/docs/guide/sitemaps.md
@@ -6,6 +6,8 @@ XML sitemaps help search engines discover and index pages on your wiki. Canasta 
 
 - [Why sitemaps matter](#why-sitemaps-matter)
 - [Generating sitemaps](#generating-sitemaps)
+  - [Redirect pages](#redirect-pages)
+  - [Filtering namespaces](#filtering-namespaces)
 - [Removing sitemaps](#removing-sitemaps)
 - [Background refresh](#background-refresh)
 - [File storage](#file-storage)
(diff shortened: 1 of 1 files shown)


yaronkoren (approved):

cicalese:
NS_MAIN and NS_HELP are integer constants, not strings. See https://www.mediawiki.org/wiki/Manual:Namespace_constants.

PR #415, opened 2026-02-26
Add test for list command

### LINKED ISSUE #505: Add tests for the list command
## Context

The `List()` function in `cmd/list/list.go` has no test coverage. It handles two code paths:
- Displaying all registered installations with their wiki details
- Cleaning up stale entries whose installation directories no longer exist (`--cleanup` flag)

The underlying `config.ListAll()` also annotates missing installation paths with `[not found]`.

## What to test

- Normal listing with one or more registered installations and wikis
- The `--cleanup` flag removing stale entries (installation directory doesn't exist)
- The `[not found]` annotation for missing installation paths without cleanup

Ref #241

Deepan200201:
Add tests for the `List()` function in `cmd/list/list.go`:

- `TestList` — registers an installation with two wikis and verifies the table output contains all expected fields
- `TestListCleanup` — registers an installation with a missing directory, calls `List` with `cleanup=true`, and verifies the stale entry is removed
- `TestListNotFound` — registers an installation with a missing directory and verifies the `[not found]` annotation appears in the output

Fixes #505

### DIFF
--- a/cmd/list/list_test.go
+++ b/cmd/list/list_test.go
@@ -1,0 +1,86 @@
+package list
+
+import (
+	"bytes"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/CanastaWiki/Canasta-CLI/internal/config"
+	"github.com/CanastaWiki/Canasta-CLI/internal/farmsettings"
+)
+
+func TestList(t *testing.T) {
+	tmpDir, err := os.MkdirTemp("", "canasta-list-test")
+	if err != nil {
+		t.Fatalf("Failed to create temp config dir: %v", err)
+	}
+	defer os.RemoveAll(tmpDir)
+
+	config.ResetForTesting(tmpDir)
+
+	installPath := filepath.Join(tmpDir, "canasta-test")
+	err = os.MkdirAll(filepath.Join(installPath, "config"), 0755)
+	if err != nil {
+		t.Fatalf("Failed to create mock config dir: %v", err)
+	}
+
+	err = farmsettings.AddWiki("testwiki", installPath, "testwiki.local", "/", "Test Wiki")
+	if err != nil {
+		t.Fatalf("Failed to add mock wiki: %v", err)
+	}
+	err = farmsettings.AddWiki("devwiki", installPath, "devwiki.local", "/", "Dev Wiki")
+	if err != nil {
+		t.Fatalf("Failed to add second mock wiki: %v", err)
+	}
+
+	installation := config.Installation{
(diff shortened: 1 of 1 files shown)


cicalese (commented):
Thanks for the contribution! The test is well-structured — good use of `ResetForTesting` for isolation and proper cleanup with `defer os.RemoveAll`.

A few suggestions:

1. **Test the `--cleanup` path**: The more interesting logic in `List()` is the cleanup of stale entries. It would be great to add a test case where the installation directory is removed before calling `List(instance, true)`, and then verify the stale entry gets cleaned up.

2. **Test the `[not found]` annotation**: `ListAll()` marks entries whose directories don't exist. Since the test always creates the directory, this path isn't exercised. A simple additional case where you remove the directory before calling `List()` would cover this.

3. **Unused `instance` parameter**: `List()` receives a `config.Installation` but never uses it (the test passes an empty struct). This isn't a problem with your test — it's a quirk of the function signature — but worth noting.

4. **Minor**: The stdout capture via `os.Pipe` works fine for now, but can cause issues if tests run in parallel. Not a blocker for this PR, just something to be aware of.

Would you be up for adding a cleanup test case? That would make this a really solid addition.

Deepan200201:
@cicalese Thanks for the review. I have added the additional tests too now.

cicalese (commented):
Thanks for adding the cleanup and not-found tests — these cover the paths I asked about nicely.

A few suggestions before merging:

1. **Use `t.TempDir()` instead of `os.MkdirTemp` + `defer os.RemoveAll`** — All three tests manually create and clean up temp dirs. `t.TempDir()` handles cleanup automatically and is the pattern used throughout the codebase (e.g., `extensionsskins_test.go`, `canasta_test.go`). Since you're already touching the file, it'd be a good time to update the original `TestList` as well.

2. **Reset global config state after each test** — Each test calls `config.ResetForTesting(tmpDir)` but doesn't restore it afterward. If other tests in the same package run later, the config package's global state could leak. Add a `t.Cleanup` call, e.g.:
   ```go
   config.ResetForTesting(tmpDir)
   t.Cleanup(func() { config.ResetForTesting("") })
   ```

3. **Minor: `TestListCleanup` could also check the table output** — The test verifies the "Removed stale entry" message and confirms the entry is gone via `GetAll()`, which is good. Optionally, you could also check that the stale entry doesn't appear in the printed table, but this isn't a blocker since `GetAll()` already covers it.

None of these are functional bugs — the test logic is correct. Just style/hygiene improvements to match the rest of the codebase.

Deepan200201:

cicalese (approved):

PR #416, opened 2026-02-26
Clean up Cobra command patterns

### LINKED ISSUE #409: Remove duplicate verbose flag from restart command
**Parent:** #397

`cmd/restart/restart.go` defines its own local `--verbose` flag (line 18, 53) which shadows the global `--verbose` flag defined in `root.go`. The `start` and `stop` commands don't do this — they rely on the global flag. Remove the local flag from restart to match.

### LINKED ISSUE #410: Standardize command initialization patterns across cmd packages
**Parent:** #397

Command packages use inconsistent initialization patterns:
- Some use package-level globals set in `init()` or closures (`cmd/extension`, `cmd/skin`, `cmd/backup`)
- Others use local variables inside `NewCmdCreate()` (`cmd/stop`, `cmd/create`)

Should pick one pattern and apply consistently. Local variables in `NewCmdCreate()` are preferred as they're easier to test and avoid shared mutable state.

### LINKED ISSUE #411: Standardize Cobra constructor naming convention
**Parent:** #397

Cobra command constructors use inconsistent naming:
- `NewCmdCreate()` — used by top-level commands in `extension.go`, `skin.go`, etc.
- `enableCmdCreate()`, `disableCmdCreate()`, `listCmdCreate()` — used by subcommands

Should standardize on one pattern (e.g., `NewCmdEnable()`, `NewCmdDisable()`, `NewCmdList()`).

### LINKED ISSUE #413: Fix verbose logging message in root.go
**Parent:** #397

`root.go:43` prints "Setting verbose" unconditionally in `PersistentPreRun`, even when verbose is false. Should either only print when verbose is true, or include the value in the message.

cicalese:
## Summary

- Standardized Cobra command patterns across all packages — consistent PersistentPreRunE, flag registration, and help text formatting.

Closes #409
Closes #410
Closes #411
Closes #413

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes


yaronkoren:
Looks good! What's better about newXYZ() vs. XYZCreate()? (Besides being more grammatically correct?) Is that a Go convention?

cicalese:
> Looks good! What's better about newXYZ() vs. XYZCreate()? (Besides being more grammatically correct?) Is that a Go convention?

Actually, Go/Cobra convention and consistency. We were using several different naming approaches, apparently.

yaronkoren:
That's true - sometimes it was NewXYZCreate().

cmd/backup/delete.go
```diff
@@ -16,20 +19,16 @@ func deleteCmdCreate() *cobra.Command {
 may still exist until a prune is run on the repository.`,
 		Example: `  canasta backup delete -i myinstance -s abc123`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return deleteSnapshot()
+			output, err := runBackup(*orch, instance.Path, *envPath, nil, "-r", *repoURL, "forget", snapshot)
```
yaronkoren, line 22: This is a pretty big functional change, no - to run backup as part of a delete? Does it belong in this "cleanup" PR?

yaronkoren (commented):
Just one comment, I think - see inline.

cmd/backup/delete.go
cicalese: It's doing the same thing. The old deleteSnapshot() function ran "forget". Now we're eliminating that function and just running it here. The command is `canasta backup delete`, which is explicitly asking for an old backup to be deleted (forgotten).

yaronkoren: Oh, I get it... runBackup() does not necessarily run a backup.

yaronkoren (approved):

PR #417, opened 2026-02-26
Deduplicate extension/skin commands and consolidate Remove functions

cicalese:
## Summary

- **Extension/skin command deduplication (#402):** Moved Cobra command construction (parent, enable, disable, list) into parameterized builder functions in `internal/extensionsskins/commands.go`. The `cmd/extension/` and `cmd/skin/` packages are now single-file thin wrappers that define their `Item` and call the builders. Deleted 6 duplicate files.
- **Remove function consolidation (#403):** Extracted a `removeWikiDir` helper in `internal/canasta/canasta.go` and rewrote `RemoveSettings`, `RemoveImages`, and `RemovePublicAssets` as thin wrappers around it.

Depends on #416.

Closes #402
Closes #403

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes
- [x] `canasta extension --help` / `canasta skin --help` output reads naturally
- [x] `canasta extension enable --help` / `canasta skin enable --help` show correct examples
- [x] `canasta extension list -i wikifarm` — output matches previous behavior
- [x] `canasta skin list -i wikifarm` — output matches previous behavior

PR #420, opened 2026-02-26
Deduplicate extension/skin commands

### LINKED ISSUE #402: Deduplicate extension/skin enable/disable commands
**Parent:** #397

The following four files have nearly identical `RunE` bodies, differing only in the type constant (extension vs skin):
- `cmd/extension/enable.go`
- `cmd/extension/disable.go`
- `cmd/skin/enable.go`
- `cmd/skin/disable.go`

Extract the shared logic into a common function in `internal/extensionsskins/` that accepts the type as a parameter.

cicalese:
## Summary

- Move shared command-building logic (list, enable, disable) into `internal/extensionsskins/commands.go`
- `cmd/extension` and `cmd/skin` become thin wrappers passing an `Item` config
- Add `CmdName`, `Plural`, and `ExampleNames` fields to the `Item` struct so help text is generated dynamically

Split from #417.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes
- [x] Verify `canasta extension list/enable/disable` and `canasta skin list/enable/disable` behave the same as before

Closes #402

### DIFF
--- a/cmd/extension/extension.go
+++ b/cmd/extension/extension.go
@@ -1,54 +1,18 @@
 package extension
 
 import (
-	"log"
-	"os"
-
 	"github.com/spf13/cobra"
 
-	"github.com/CanastaWiki/Canasta-CLI/internal/canasta"
-	"github.com/CanastaWiki/Canasta-CLI/internal/config"
 	"github.com/CanastaWiki/Canasta-CLI/internal/extensionsskins"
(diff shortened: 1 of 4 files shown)


yaronkoren (approved):
Much more streamlined!

PR #422, opened 2026-02-26
Standardize logging across command packages

### LINKED ISSUE #408: Standardize logging: pick one approach and use consistently
**Parent:** #397

The codebase mixes multiple logging approaches with no clear rationale:
- `log.Fatal()` — used in `create.go`, `start.go`, `extension.go`, `add.go`, `remove.go`
- `logging.Fatal()` — used in `root.go`
- `fmt.Print()` / `fmt.Println()` — used in various places for user-facing output
- `logging.Print()` — used for verbose/debug output
- `log.Printf()` / `log.Println()` — used in `extensionsskins.go`

Should standardize: use `logging.*` for debug/verbose output, `fmt.*` for user-facing output, and pick one fatal approach.

cicalese:
## Summary

- Replace `log.Print/Printf/Println` with `fmt.*` equivalents in `internal/extensionsskins/extensionsskins.go` — these are user-facing messages that should always print, not go through the log package
- Replace `log.Fatal` with `logging.Fatal` across all command packages so fatal errors print without a timestamp prefix
- Update imports accordingly (swap `"log"` for `"…/logging"` or remove `"log"` where `logging` was already imported)

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes
- [x] `grep -rn 'log\.Fatal\|log\.Print' --include='*.go' | grep -v tools/docgen | grep -v internal/logging` returns nothing

Closes #408

### DIFF
--- a/cmd/add/add.go
+++ b/cmd/add/add.go
@@ -2,7 +2,6 @@ package add
 
 import (
 	"fmt"
-	"log"
 	urlpkg "net/url"
 	"os"
 	"strings"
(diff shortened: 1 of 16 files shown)


yaronkoren (approved):
Looks good! Thank you!

PR #436, opened 2026-02-26
Define service name constants for container orchestration

### LINKED ISSUE #428: Define service name constants for "web" and "db"
Part of #423 (code quality audit round 2).

### Problem

62 occurrences of `"web"`, `"db"` string literals are scattered across 19+ files as arguments to orchestrator methods (`ExecWithError`, `CopyTo`, `CopyFrom`, etc.). These magic strings are a typo risk and make it harder to refactor service names.

Note: `"cache"` is not currently used anywhere.

`internal/mediawiki/mediawiki.go` already defines a local `dbServer = "db"` constant, but this is package-private and not used elsewhere.

### Fix

Define exported constants in `internal/orchestrators/`:

```go
const (
    ServiceWeb = "web"
    ServiceDB  = "db"
)
```

Replace all hardcoded `"web"` and `"db"` string literals in orchestrator method calls across the codebase. Remove the local `dbServer` constant in mediawiki.go.

### Files affected

Approximately 19 files with `"web"` or `"db"` literals in orchestrator calls, including:
- `internal/mediawiki/mediawiki.go`
- `internal/extensionsskins/extensionsskins.go`
- `internal/orchestrators/orchestrator.go`
- `cmd/backup/create.go`, `cmd/backup/restore.go`
- `cmd/export/export.go`
- `cmd/remove/remove.go`, `cmd/delete/delete.go`
- `cmd/upgrade/upgrade.go`
- `cmd/maintenanceUpdate/` (multiple files)
- `cmd/sitemap/` (multiple files)

cicalese:
## Summary

- Define `ServiceWeb` and `ServiceDB` constants in `internal/orchestrators/`
- Replace all hardcoded `"web"` and `"db"` string literals used as service name arguments across 16 files
- Remove the package-private `dbServer` constant from `mediawiki.go` that this supersedes

This is a purely mechanical refactoring — no behavioral change. Eliminates typo risk and makes service names easy to refactor if container service names ever change.

## Test plan

Mechanical search-and-replace of string literals with constants. Verified by:
- [x] `go build ./...` — compiles cleanly
- [x] `go test ./...` — all tests pass
- [x] `grep` confirms no remaining `"web"` or `"db"` literals in orchestrator method calls

Fixes #428


internal/mediawiki/mediawiki.go
```diff
@@ -218,8 +217,8 @@ func InstallOne(installPath, id, domain, admin, adminPassword, dbuser, workingDi
 		installCmd = "php maintenance/install.php"
 	}
 	command := fmt.Sprintf("%s --skins='Vector' --dbserver=%s --dbname=%s --confpath=%s --scriptpath=%s --server=%s --installdbuser='%s' --installdbpass=%s --dbuser='%s' --dbpass=%s --pass=%s %s %s",
-		installCmd, dbServer, orchestrators.ShellQuote(id), confPath, scriptPath, orchestrators.ShellQuote("https://"+domain), installdbuser, orchestrators.ShellQuote(installdbpass), dbuser, orchestrators.ShellQuote(dbpass), orchestrators.ShellQuote(adminPassword), orchestrators.ShellQuote(id), orchestrators.ShellQuote(admin))
-	output, err := orch.ExecWithError(installPath, "web", command)
+		installCmd, orchestrators.ServiceDB, orchestrators.ShellQuote(id), confPath, scriptPath, orchestrators.ShellQuote("https://"+domain), installdbuser, orchestrators.ShellQuote(installdbpass), dbuser, orchestrators.ShellQuote(dbpass), orchestrators.ShellQuote(adminPassword), orchestrators.ShellQuote(id), orchestrators.ShellQuote(admin))
```
yaronkoren, line 220: Is this correct? Doesn't dbServer hold an actual server name?

yaronkoren: Oh, I just saw the PR summary - I guess not.

yaronkoren: Actually, let me make sure I understand this - the MW command install.php takes in a "--dbserver" flag, and that's getting set to the string "db"? Is that correct, and if so, is it a coincidence that this string is the same as the string "db" that's used as an internal flag in this code, or is that on purpose?

cicalese: The name of the running container (AKA service) that holds the database is "db". As far as MediaWiki is concerned, it is the name (as in the domain name) of the server it needs to contact for database actions. So, regardless of what the host's actual server name/domain name is, inside the container network, the database server is always known as "db".

cicalese: Basically, dbServer was an earlier attempt to use a constant for the same thing, but only locally to this file.

yaronkoren: Okay, I sort of understand. I don't know... this is a minor point, but it might be "too clever" to use the Orchestrator.ServiceDB value for this dbServer name. It seems to me like two different strings, both of which are "db", but they don't have to be the same.

cicalese: They have to be the same. Service is what docker/kubernetes call the containers. The service name is the name you use to address that service, which is what you would pass to MediaWiki so it can communicate with it. Different names for the same thing.

yaronkoren: Oh... okay. I thought "db" and "web" were just sort of IDs in the code. Thanks for the explanation.

yaronkoren (approved):

PR #452, opened 2026-02-27
Remove --all flag from maintenance commands, default to all wikis

### LINKED ISSUE #451: Remove --all flag from maintenance commands, default to all wikis
The `canasta maintenance update` and `canasta maintenance extension` commands require `--all` to run on every wiki in a farm. Without it, multi-wiki farms produce an error: "multiple wikis found; use --wiki=<id> or --all".

This is unnecessary ceremony. The simpler behavior is:
- `--wiki=X` → run for wiki X
- No `--wiki` → run for all wikis

Remove the `--all` persistent flag and default to all wikis when `--wiki` is not specified.

cicalese:
## Summary
- Remove the `--all` persistent flag from `canasta maintenance`
- When `--wiki` is not specified, maintenance commands now run on all wikis by default
- Simplifies the interface: `--wiki=X` targets one wiki, omitting it targets all
- Update help text, examples, and tests accordingly

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./...` passes (29 maintenance tests)
- [x] `canasta maintenance --help` shows no `--all` flag
- [x] `canasta maintenance update --help` describes default-to-all behavior

Fixes #451

### DIFF
--- a/cmd/maintenanceUpdate/extension.go
+++ b/cmd/maintenanceUpdate/extension.go
@@ -17,7 +17,7 @@ import (
 // wikiArgRe matches --wiki=value or --wiki value in a script argument string.
 var wikiArgRe = regexp.MustCompile(`(?:^|\s)--wiki[=\s](\S+)`)
 
-func newExtensionCmd(instance *config.Installation, wiki *string, all *bool) *cobra.Command {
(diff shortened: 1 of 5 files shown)


yaronkoren (approved):
Awesome! Streamlined and standardized.

PR #463, opened 2026-02-27
Switch extension/skin enable/disable to YAML-based config

### LINKED ISSUE #462: Switch extension/skin enable/disable to YAML-based config
Currently, `canasta extension enable Foo` creates a PHP file (`Foo.php`) containing `wfLoadExtension('Foo')` in the settings directory. `disable` deletes that file. This is fragile: user edits are lost on disable, and user-created PHP files could be accidentally deleted.

Switch to a shared `config.yaml` in each settings directory (global and per-wiki). The CLI manages this YAML file; CanastaBase reads it and calls `wfLoadExtension`/`wfLoadSkin`. User PHP files in the same directory are untouched.

```yaml
# DO NOT EDIT - managed by Canasta CLI
extensions:
- VisualEditor
- Cite
skins:
- Vector
```

**Paths:**
- Global: `config/settings/global/config.yaml`
- Per-wiki: `config/settings/wikis/{wiki}/config.yaml`

The YAML format is intentionally compatible with MediaWiki's [YAML settings file format](https://www.mediawiki.org/wiki/Manual:YAML_settings_file_format) used by `SettingsBuilder::loadFile()`, so a future migration to that API would require no CLI-side changes.

Requires a companion change in CanastaBase.

cicalese:
## Summary

- Replace the PHP-file-per-extension approach with a shared `main.yaml` in each settings directory
- CLI writes YAML on the host filesystem; CanastaBase reads it via `SettingsBuilder::loadFile()`
- `Enable`/`Disable`/`CheckEnabled` no longer need the orchestrator (no container exec)
- Add 13 new tests covering YAML read/write, enable, disable, and check operations
- Update documentation to describe YAML mechanism and settings file loading order

Requires companion CanastaBase change: CanastaWiki/CanastaBase#118.

## Test plan

- [x] `go build ./...` / `go vet ./...` / `go test ./...` all pass
- [x] `canasta extension enable VisualEditor -i wikifarm` creates `config/settings/global/main.yaml`
- [x] `canasta skin enable Timeless -i wikifarm` adds Timeless under skins in same file
- [x] `canasta extension disable VisualEditor -i wikifarm` removes from YAML
- [x] `canasta extension enable Cite -i wikifarm -w wiki1` creates per-wiki main.yaml
- [x] Old-style PHP files still work alongside new YAML mechanism
- [x] Extensions/skins confirmed loaded in MediaWiki via API with CanastaBase changes deployed

Fixes #462


internal/extensionsskins/extensionsskins.go
```diff
@@ -2,14 +2,29 @@ package extensionsskins
...
 )
 
+// configYAML represents the YAML config file managed by the CLI.
+type configYAML struct {
+	Extensions []string `yaml:"extensions,omitempty"`
+	Skins      []string `yaml:"skins,omitempty"`
+}
+
+const configFileName = "main.yaml"
+const configHeader = "# This file is managed by Canasta, which adds and removes extensions and\n# skins from it. You may edit it, but Canasta may overwrite your changes.\n"
```
yaronkoren: What does "overwrite" here mean - that lines could be deleted?

internal/extensionsskins/extensionsskins.go
```diff
@@ -58,93 +125,84 @@ func CheckInstalled(name string, instance config.Installation, orch orchestrator
...
 
-	output, err = orch.ExecWithError(instance.Path, orchestrators.ServiceWeb, fmt.Sprintf(`cat %s`, filePath))
-	if err != nil {
-		return "", fmt.Errorf("failed to read settings for %s %q: %s", constants.Name, name, output)
-	}
-	if !slices.Contains(strings.Split(output, "\n"), "// This file was generated by Canasta") {
-		return "", fmt.Errorf("%s %s was not generated by Canasta cli", constants.Name, name)
+	slice := getSlice(&cfg, constants)
+	if !slices.Contains(*slice, name) {
+		return "", fmt.Errorf("%s %s is not enabled", constants.Name, name)
```
yaronkoren: This message is misleading, if I understand the code correctly - if it's just checking against they YAML file(s), it should probably just say something like "An extension/skin named X was not found in the standard main.yaml file, so if it is enabled, it cannot be disabled via this command". Maybe there's a shorter way of saying that. :)

internal/extensionsskins/extensionsskins.go
cicalese: Lines could be added or deleted. Changes made by a user could be lost. Hopefully the risk is minimal, but it would be better for users to make their changes in other files. I want them to get the sense "use at your own risk".

internal/extensionsskins/extensionsskins.go
cicalese: I'm changing it to:

%s %s is not listed in %s so it was not removed

internal/extensionsskins/extensionsskins.go
yaronkoren: I don't know... this feels confusing to me. There are two ways to modify the main YAML file: via the command line, and by editing it directly. Either way, the lines one adds can be removed later. So why should one get a warning and the other doesn't?

internal/extensionsskins/extensionsskins.go
yaronkoren: Sounds good. 👍

internal/extensionsskins/extensionsskins.go
cicalese: Because the edits done through `canasta extension enable` and `canasta extension disable` are limited in scope. I used to work with someone who would say, "The user could go in and write SATAN SATAN SATAN all over the file". Users need to exhibit care when they are editing the file. Hopefully the CLI is more disciplined than your typical user.

yaronkoren: Well, okay, but if they mess up the the file, making it invalid YAML, then the Canasta commands won't modify it at all - so there's even less of a chance of their "changes being overwritten".

cicalese: If the user's edits make one of the settings files, yaml or php, invalid, it will likely result in a 500 error at runtime. While the CLI is editing the file, it would probably give an error message and not edit the file, although there could be changes that would be valid yaml but not match the SettingsBuilder schema. Is your issue only with the word "overwrite"? If so, tell me what warning you would prefer that would let the user know that they're not the only ones manipulating the contents of the file and that they may lose their work.

yaronkoren: I personally don't think any warning is needed - it's the same admin(s) who are editing the file, in two different ways; nothing notable about one way vs. the other, in my opinion.

cicalese: OK, removed.

yaronkoren:
This looks great. The "config:" header is still there in the sample file in extensions-skins.md. Is that accidental, or on purpose, or it doesn't matter?

cicalese:
Thanks! I could have sworn I removed that. It is gone now. Good to go!

yaronkoren (approved):
Great, thank you! This is an awesome improvement!

PR #465, opened 2026-02-27
Run update.php automatically after enabling extensions/skins

### LINKED ISSUE #464: Run maintenance update after enabling extensions or skins
After `canasta extension enable` or `canasta skin enable`, run the maintenance update sequence (update.php, runJobs.php, and optionally rebuildData.php) so that database schema changes required by newly enabled extensions are applied automatically.

Currently, users must manually run `canasta maintenance update` after enabling extensions, which is easy to forget and can lead to errors.

Traced from CanastaWiki/Canasta-DockerCompose#67. Replaces #319, which was too far diverged from the current codebase to rebase.

cicalese:
## Summary

- Extracts reusable maintenance update functions (`RunUpdatePhp`, `RunUpdateAllWikis`) into a new `internal/maintenance` package
- Moves `GetWikiIDs` to `internal/farmsettings` where it belongs alongside the other wiki-config functions
- After enabling extensions or skins, `update.php --quick` is automatically run for all relevant wikis to apply any required database schema changes
- Adds `--skip-update` flag to bypass the automatic `update.php` run

Resolves #464
See also CanastaWiki/Canasta-DockerCompose#67

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go test ./...` all pass
- [x] `canasta extension enable VisualEditor -i <instance>` enables the extension and runs `update.php`
- [x] `canasta extension enable VisualEditor -i <instance> --skip-update` enables without running `update.php`
- [x] `canasta skin enable Timeless -i <instance>` enables the skin and runs `update.php`
- [x] `canasta maintenance update -i <instance>` still works as before
- [x] `canasta extension enable VisualEditor -i <instance> -w docs` runs `update.php` only for the specified wiki

### DIFF
--- a/internal/extensionsskins/commands.go
+++ b/internal/extensionsskins/commands.go
@@ -10,6 +10,7 @@ import (
 	"github.com/CanastaWiki/Canasta-CLI/internal/canasta"
 	"github.com/CanastaWiki/Canasta-CLI/internal/config"
 	"github.com/CanastaWiki/Canasta-CLI/internal/logging"
+	"github.com/CanastaWiki/Canasta-CLI/internal/maintenance"
 	"github.com/CanastaWiki/Canasta-CLI/internal/orchestrators"
 )
 
@@ -69,6 +70,8 @@ is shown with its enabled/disabled status.`, constants.Plural, constants.CmdName
 }
 
 func newEnableCmd(instance *config.Installation, orch *orchestrators.Orchestrator, wiki *string, constants *Item) *cobra.Command {
+	var skipUpdate bool
+
 	// Build the Use string with an appropriate argument placeholder
 	argName := strings.ToUpper(constants.CmdName)
 	useStr := fmt.Sprintf("enable %s1,%s2,...", argName, argName)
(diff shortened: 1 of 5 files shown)


yaronkoren:
Is this like three unrelated changes in one?

cicalese:
It does some cleanup to facilitate the change.

yaronkoren:
What's the connection between the main change and, say, renaming the "maintenanceUpdate" directory?

cicalese:
It was the only odd directory out in terms of naming. The connection is tenuous, but if I pull it out into a separate PR, this PR will depend on it, and GitHub doesn't do stacked PRs well, so it would require rebasing and merge conflict resolution.

internal/extensionsskins/commands.go
```diff
@@ -88,18 +91,25 @@ func newEnableCmd(instance *config.Installation, orch *orchestrators.Orchestrato
...
-for a specific wiki only.`, constants.Plural, constants.Plural, article(constants.CmdName), constants.CmdName),
+for a specific wiki only.
+
+After enabling, update.php is automatically run to apply any required
+database changes. Use --skip-update to skip this step.`, constants.Plural, constants.Plural, article(constants.CmdName), constants.CmdName),
 		Example: example,
 		Args:    cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			names := strings.Split(args[0], ",")
+			anyEnabled := false
```
yaronkoren, line 112: Is this variable needed? Can't you just check if "names" is non-empty?

yaronkoren:
Well, that's unfortunate - this patch changes 11 files, and maybe 2 of them have to do with what's in the commit summary.

cicalese:
Fair enough. I'll break it out.

internal/extensionsskins/commands.go
cicalese: Yes — `names` is always non-empty (it comes from `strings.Split` on a required arg), but not every name necessarily succeeds. If `CheckInstalled` fails for a name, it prints the error and `continue`s without enabling anything. `anyEnabled` tracks whether at least one extension was actually enabled, so we only run `update.php` when there's something to update. Without it, we'd run `update.php` even when all enables were skipped.

yaronkoren: Oh, that makes sense... the diff didn't show the lines holding the "continue".

cicalese:
The rename is now removed from this PR and is in #469. This should hopefully be ready now.

yaronkoren (approved):
This looks good enough. Thank you for this great addition!

PR #472, opened 2026-02-27
Fix db container health check stalling on restart

### LINKED ISSUE #471: DB health check stalls on restart
## Problem

After `canasta upgrade` or `canasta restart`, the db container health check never passes, which blocks the web and varnish containers from starting.

Two issues contribute:

1. **Health check variable scoping**: The health check uses `CMD` form with `${MYSQL_ROOT_PASSWORD}`, which Docker Compose interpolates on the host. But only `MYSQL_PASSWORD` is defined in `.env` — `MYSQL_ROOT_PASSWORD` only exists inside the container (mapped from `MYSQL_PASSWORD`). The variable resolves to blank, so the health check authenticates with an empty password and fails.

2. **`skip-binary-as-hex` in my.cnf**: The `mysqladmin` tool reads `my.cnf` and does not recognize the `skip-binary-as-hex` option (it's a `mysql` client-only option), causing the health check command to error with `unknown variable`.

## Expected behavior

`canasta restart` should bring all containers up without stalling.

cicalese:
Fixes #471

## Summary

- Switch db health check from `CMD` to `CMD-SHELL` so `$MYSQL_ROOT_PASSWORD` is evaluated inside the container where it's defined, not on the host where only `MYSQL_PASSWORD` exists in `.env`
- Add upgrade migration to remove `skip-binary-as-hex` from non-`[mysql]` sections of existing installations' `my.cnf` (not recognized by `mysqladmin`, breaks health check) - to fix old Canasta installations

## Test plan

- [x] Migration preserves `skip-binary-as-hex` in `[mysql]` section
- [x] Migration removes `skip-binary-as-hex` from `[client]`, `[mysqld]`, and other sections
- [x] Migration removes `skip-binary-as-hex` appearing before any section header
- [x] Migration is a no-op when `skip-binary-as-hex` is absent or file is missing
- [x] `canasta upgrade --dry-run` reports the migration without applying it
- [ ] `canasta restart` completes without the db health check stalling

### DIFF
--- a/cmd/upgrade/upgrade.go
+++ b/cmd/upgrade/upgrade.go
@@ -302,6 +302,15 @@ func runMigration(installPath string, orch orchestrators.Orchestrator, dryRun bo
 		changed = true
 	}
 
+	// Step 7: Remove skip-binary-as-hex from my.cnf (breaks mysqladmin health check)
+	mycnfChanged, err := removeSkipBinaryAsHex(installPath, dryRun)
+	if err != nil {
+		return false, err
+	}
+	if mycnfChanged {
+		changed = true
+	}
+
 	if !changed {
 		fmt.Println("No config migrations needed.")
 	} else if dryRun {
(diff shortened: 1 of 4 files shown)


yaronkoren:
Does this conflict with #375? If Canasta switched to using MariaDB, would any of this be necessary?

cicalese:
It will need a rebase, but unfortunately, the migration code would still be necessary, because all existing Canasta users have those problematic lines in their my.cnf files. The advantage of switching to MariaDB for Canasta is that this directive is no longer necessary - a minor concern for most users, but a big pain for us.

yaronkoren:
This approach seems pretty extreme... is the "skip-binary-as-hex = true" setting not actually useful? And is there really no way to get mysqladmin to ignore this setting? Why, actually, is it looking at my.cnf at all, if that's a client-only configuration file?

cicalese:
It is now updated to leave "skip-binary-as-hex = true" in the "[mysql]" section, which is the only section where it is useful.

yaronkoren:
So is this a fix only for older Canasta installations?

cicalese:
Yes

yaronkoren:
Alright, that makes sense. Could the commit message just be changed to make that clearer?

cicalese:
Yes, the message is updated.

yaronkoren (approved):
Good enough. Thanks!

PR #504, opened 2026-03-02
Added tests for GenerateAndSaveSecretKey

### LINKED ISSUE #496: Add tests for GenerateAndSaveSecretKey()

`GenerateAndSaveSecretKey()` has two branches: skip generation if `MW_SECRET_KEY` is already set in the env file, or generate a new key and save it. Neither branch is tested.

Test cases:
- When `MW_SECRET_KEY` is already in the `.env` file, the function should not modify the file
- When `MW_SECRET_KEY` is not set, the function should generate a 64-character hex string and save it

Relates to #241

Deepan200201:
Fixes https://github.com/CanastaWiki/Canasta-CLI/issues/496

### DIFF
--- a/internal/canasta/canasta_test.go
+++ b/internal/canasta/canasta_test.go
@@ -1,6 +1,8 @@
 package canasta
 
 import (
+	"bytes"
+	"encoding/hex"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


cicalese:
Thank you for your contribution! Overall, the PR is solid and covers both branches as described in issue #496. Two tests, clean structure, correct assertions. A few suggestions:

1. Missing `.env` file test — `TestGenerateAndSaveSecretKey_GeneratesNew` creates a `.env` with `KEY1=value1` but no `MW_SECRET_KEY`. The function will call `GetEnvVariable` then `SaveEnvVariable`. It would be worth adding a case where the `.env` file doesn't exist at all — that would exercise the error path on `GetEnvVariable`.
2. "Already set" test could be stronger — The test checks that the key value is unchanged, which is good. But it doesn't verify that the rest of the file is also unchanged. A simple `bytes.Equal` comparison of the file contents before and after would make the "no modification" assertion more robust, as mentioned in the issue ("should not modify the file").
3. Hex validation missing — `TestGenerateAndSaveSecretKey_GeneratesNew` checks `len(key) != 64` but doesn't verify the key is valid hex. A quick regex or `hex.DecodeString` check would confirm the generated key matches the expected format from `generateSecretKey()`.
4. PR description — "Added tests for GenerateAndSaveSecretKey" is fine but the PR description should reference the issue: "Fixes #496" or "Ref #496".

Deepan200201:

cicalese (approved):

Issue #241, opened 2026-02-12
Improve test coverage across packages


cicalese:
## Context

PR #232 added initial unit tests for several internal packages (`canasta`, `config`, `farmsettings`, `git`, `orchestrators`) along with a CI workflow that runs `go build`, `go vet`, and `go test` on all PRs.

## Current coverage

The following packages have tests:
- `internal/canasta` — env variables, password generation, database path validation, secret key generation
- `internal/config` — add, get, delete, update, exists operations
- `internal/farmsettings` — wiki ID validation, wikis.yaml read/write, add/remove wiki
- `internal/git` — FetchAndCheckout file categorization
- `internal/orchestrators` — Exec, StopAndStart, EnsureRunning, ImportDatabase (with mocked execute)

## Packages with no tests

- `internal/devmode`
- `internal/execute`
- `internal/extensionsskins`
- `internal/imagebuild`
- `internal/logging`
- `internal/mediawiki`
- `internal/selfupdate`
- `internal/spinner`
- `internal/system`
- `cmd/*` (all command packages)

## Goals

- [ ] Add tests for remaining internal packages
- [ ] Add integration-style tests for key command workflows (create, delete, etc.)
- [ ] Add coverage reporting to CI
- [ ] Establish a coverage baseline and track regressions

cicalese:
## Additional untested code paths identified (March 2026 audit)

### High priority

1. **`extractSecretKey()`** in `cmd/upgrade/upgrade.go` — 3 major code paths: reading from a dedicated file, regex-parsing from LocalSettings.php, and fallback generation. Complex string/regex logic with no coverage.

2. **`migrateDirectoryStructure()`** in `cmd/upgrade/upgrade.go` — handles migration from the old flat layout to the new directory structure. File moves, directory creation, conditional logic — all untested.

3. **`syncComposeProfiles()`** in `internal/orchestrators/compose.go` — reads/writes `COMPOSE_PROFILES` in `.env` based on feature flags (`CANASTA_ENABLE_ELASTICSEARCH`, `CANASTA_ENABLE_OBSERVABILITY`). Has append/remove logic that could easily regress.

4. **`getComposeImages()`** in `internal/orchestrators/compose.go` — parses `docker compose config --images` output line-by-line. Brittle string parsing that could break if the output format changes.

[...]

PR #506, opened 2026-03-02
tests: add table-driven tests for getRepoURL

### LINKED ISSUE #489: Add tests for getRepoURL()
**File:** `cmd/backup/backup.go`
**Test file:** `cmd/backup/backup_test.go`

`getRepoURL()` takes a map of environment variables and returns the backup repository URL. It has three code paths, none of which are tested:

1. When `RESTIC_REPOSITORY` is set, it should be returned
2. When `RESTIC_REPOSITORY` is not set but `RESTIC_REPO` is, `RESTIC_REPO` should be returned
3. When neither is set, it should construct a URL from `AWS_S3_API` and `AWS_S3_BUCKET` in the format `s3:<api>/<bucket>`

Follow the table-driven test pattern already used in the file.

Relates to #241

lakxy:
Adds test coverage for `getRepoURL()` in `cmd/backup/backup_test.go`

### Changes

- Added TestGetRepoURL using the existing table-driven pattern in the file

### Test cases covered

- `RESTIC_REPOSITORY` set, takes precedence over all other vars
- `RESTIC_REPO` set, used when `RESTIC_REPOSITORY` is absent
- AWS S3 vars set, constructs `s3:<api>/<bucket>`
- All vars empty, graceful fallback to `s3:/`

### Testing

```bash
go test ./cmd/backup/ -run TestGetRepoURL -v
```

test result
image

### Related

Closes #489
Relates to #241

### DIFF
--- a/cmd/backup/backup_test.go
+++ b/cmd/backup/backup_test.go
@@ -184,3 +184,56 @@ func TestRestorePreservesDBPasswords(t *testing.T) {
 		t.Errorf("expected MW_SITE_SERVER from backup to be preserved, got:\n%s", content)
 	}
 }
+
+func TestGetRepoURL(t *testing.T) {
+	for _, tc := range []struct {
+		name string
+		env  map[string]string
+		want string
+	}{
+		{
+			name: "RESTIC_REPOSITORY is set",
+			env: map[string]string{
+				"RESTIC_REPOSITORY": "rest:http://repo.example/repo",
+				"RESTIC_REPO":       "rest:http://repo.example/other",
+				"AWS_S3_API":        "https://s3.example",
+				"AWS_S3_BUCKET":     "bucket",
+			},
+			want: "rest:http://repo.example/repo",
+		},
+		{
+			name: "RESTIC_REPO is set",
+			env: map[string]string{
+				"RESTIC_REPO":   "rest:http://repo.example/repo",
+				"AWS_S3_API":    "https://s3.example",
+				"AWS_S3_BUCKET": "bucket",
+			},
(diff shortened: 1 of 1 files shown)


cicalese (commented):
Thank you for your contribution! Clean, well-structured PR that covers all three code paths with the correct table-driven pattern.

A few minor suggestions:

1. **"All vars empty" case includes explicit empty strings** — The test sets all four keys to `""`. This works, but there's a subtle difference: in the function, `RESTIC_REPOSITORY` is checked with `ok && val != ""`, so setting the key to `""` exercises the `val != ""` path. A map with no keys at all would exercise the `ok` path (map lookup returns zero value). Worth adding an additional case with an empty map `map[string]string{}` to cover that — the expected result would also be `"s3:/"`.

2. **Missing case: only one S3 var set** — What happens when `AWS_S3_API` is set but `AWS_S3_BUCKET` is missing (or vice versa)? The function would produce `"s3:https://s3.example/"` or `"s3:/bucket"`. Not necessarily a bug, but documenting this behavior in a test case makes the contract explicit.

3. **No test for `RESTIC_REPO` precedence over S3** — The `RESTIC_REPO` case also includes S3 vars, which is good. But consider adding a case where `RESTIC_REPOSITORY` is explicitly `""` and `RESTIC_REPO` is set, to verify the empty-string check on `RESTIC_REPOSITORY` correctly falls through.

lakxy:
Hi @cicalese please review. I have addressed the review and tested it too.

cicalese (approved):

PR #512, opened 2026-03-02
Test: add tests for DeleteEnvVariable 

### LINKED ISSUE #492: Add tests for DeleteEnvVariable()

`DeleteEnvVariable()` removes a key from a `.env` file and returns an error if the key is not found. It has no tests.

Test cases to cover:
- Removing an existing key from a file with multiple keys
- Removing a key that doesn't exist (should return an error)
- Removing the only key in the file

Relates to #241

lakxy:
Adds test coverage for `DeleteEnvVariable()` in `internal/canasta/canasta_test.go`, addressing #492.

### Changes

- Added `TestDeleteEnvVariable` following the existing `TestSaveEnvVariable` pattern

### Test cases covered

- Existing key removed surrounding keys remain intact
- Non-existent key returns an error
- Only key in file results in an empty map

### Testing
```bash
go test ./internal/canasta/ -run TestDeleteEnvVariable -v
```

### Test Result
image

### Related

Closes #492  
Relates to #241

### DIFF
--- a/internal/canasta/canasta_test.go
+++ b/internal/canasta/canasta_test.go
@@ -67,18 +67,15 @@ func TestSaveEnvVariable(t *testing.T) {
 	dir := t.TempDir()
 	envPath := filepath.Join(dir, ".env")
 
-	// Start with a file
 	initial := "KEY1=old\nKEY2=keep\n"
 	if err := os.WriteFile(envPath, []byte(initial), 0644); err != nil {
 		t.Fatal(err)
 	}
 
-	// Update existing key
 	if err := SaveEnvVariable(envPath, "KEY1", "new"); err != nil {
 		t.Fatalf("SaveEnvVariable() error = %v", err)
 	}
 
-	// Add new key
 	if err := SaveEnvVariable(envPath, "KEY3", "added"); err != nil {
 		t.Fatalf("SaveEnvVariable() error = %v", err)
 	}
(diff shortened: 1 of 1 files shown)


cicalese:
Thanks for your contribution! The test coverage is good — all three scenarios from the issue are covered (delete existing key, non-existent key error, delete only key).

A couple of things:

1. **Removed comments in `TestSaveEnvVariable`**: I noticed the comments "Start with a file", "Update existing key", and "Add new key" were removed. Was there a reason for that? Those comments help readers follow the test flow and I'd prefer to keep them.

2. **Table-driven subtests**: The three test scenarios are sequential within a single function, which means if the first one fails the others won't run. Could you restructure this as table-driven subtests (using `t.Run`), consistent with the pattern used elsewhere in this file? Each scenario would get its own fresh `.env` file and run independently.

Could you update the PR to address these? Thanks!

lakxy:
> Thanks for your contribution! The test coverage is good — all three scenarios from the issue are covered (delete existing key, non-existent key error, delete only key).
> 
> A couple of things:
> 
> 1. **Removed comments in `TestSaveEnvVariable`**: I noticed the comments "Start with a file", "Update existing key", and "Add new key" were removed. Was there a reason for that? Those comments help readers follow the test flow and I'd prefer to keep them.
> 2. **Table-driven subtests**: The three test scenarios are sequential within a single function, which means if the first one fails the others won't run. Could you restructure this as table-driven subtests (using `t.Run`), consistent with the pattern used elsewhere in this file? Each scenario would get its own fresh `.env` file and run independently.
> 
> Could you update the PR to address these? Thanks!

Thank you for the review, It is my mistake I removed the unrelated code, totally an accident, thank you for pointing it out.
2. does make sense I will refactorn it.

lakxy:
Hi @cicalese I have refactored the code as per review Please review!

cicalese (commented):
Nice clean test — follows the existing pattern well.

cicalese (commented):
One suggestion: consider adding a test case for a key that's a prefix of another key, e.g. deleting `"KEY"` when the file contains `KEY1=value1`. This would verify that the `HasPrefix(line, key+"=")` check in `DeleteEnvVariable` correctly avoids matching `KEY1=value1` when asked to delete `KEY`.

lakxy:
> One suggestion: consider adding a test case for a key that's a prefix of another key, e.g. deleting `"KEY"` when the file contains `KEY1=value1`. This would verify that the `HasPrefix(line, key+"=")` check in `DeleteEnvVariable` correctly avoids matching `KEY1=value1` when asked to delete `KEY`.

Testcase added, Please review!

cicalese (approved):
LGTM. Thank you for your contribution!

PR #515, opened 2026-03-02
Fix MySQL 8.0 migration failing due to volume lock

### LINKED ISSUE #514: MySQL 8.0 → MariaDB migration fails with volume lock error
## Problem

`canasta upgrade` fails during the MySQL 8.0 → MariaDB migration with:

```
Unable to lock ./ibdata1 error: 11
```

The temporary `mysql:8.0` container started for the database dump cannot acquire the InnoDB lock because the existing db container is still running and holding it.

## Cause

`migrateMySQL8ToMariaDB()` is called before `orch.Stop()` in the upgrade flow, so the db container still has the data volume locked when the temporary MySQL 8.0 container tries to start against it.

## Expected behavior

The migration dump should happen after the containers are stopped so the data volume is available.

cicalese:
## Summary
- Split `migrateMySQL8ToMariaDB()` into two phases: `detectMySQL8Data()` (read-only, safe while containers run) and `dumpMySQL8Data()` (called after `orch.Stop()`)
- Detection still happens early for dry-run reporting
- Dump now happens after containers are stopped so the data volume is unlocked

The migration was failing with `Unable to lock ./ibdata1` because the temporary `mysql:8.0` container tried to start while the existing db container still held the InnoDB lock.

Fixes #514

### DIFF
--- a/cmd/upgrade/mysql_to_mariadb.go
+++ b/cmd/upgrade/mysql_to_mariadb.go
@@ -22,36 +22,29 @@ func composeVolumeName(installPath, volume string) string {
 	return project + "_" + volume
 }
 
-// migrateMySQL8ToMariaDB detects whether the MySQL data volume contains
-// MySQL 8.0 binary data (mysql.ibd) and, if so, dumps all user databases
-// using a temporary mysql:8.0 container. Returns the path to the host-side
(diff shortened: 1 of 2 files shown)


yaronkoren (approved):
Looks like better function names, among other improvements!

PR #518, opened 2026-03-02
Fix upgrade recreating deleted files and Caddyfile.custom migration

### LINKED ISSUE #517: Upgrade recreates deleted files and fails to migrate Caddyfile.custom
## Problems

Two issues found during real-world upgrade from a pre-v1.107.0 installation:

### 1. Caddy restart loop after upgrade

After `canasta upgrade`, Caddy enters a restart loop with:
```
Error: adapting config using caddyfile: File to import not found: /etc/caddy/Caddyfile.custom
```

The Caddyfile still references the old `Caddyfile.custom` name (renamed to `Caddyfile.site` in #297). The migration in `migrateCaddyFiles()` only checks whether `Caddyfile.site` and `Caddyfile.global` exist — it doesn't check whether the Caddyfile itself still references the old name. Since `UpdateStackFiles` creates `Caddyfile.site` from the template before migrations run, the migration thinks everything is fine.

### 2. Upgrade recreates user-deleted files

`Vector.php` and `CanastaFooterIcon.php` are recreated by `UpdateInstallationTemplate()` during upgrade, even if the user intentionally deleted them. The current no-clobber logic only skips if the file exists — it recreates missing files.

## Expected behavior

1. The migration should rename `Caddyfile.custom` → `Caddyfile.site` and rewrite the Caddyfile if it references the old name.
2. User-editable template files should be skipped entirely during upgrade.

cicalese:
## Summary
- Handle `Caddyfile.custom` → `Caddyfile.site` rename during upgrade migration: rename the old file, detect stale references in the Caddyfile, and rewrite it
- Skip user-editable template files entirely during upgrade so intentionally deleted files (e.g., `Vector.php`, `CanastaFooterIcon.php`) are not recreated

## Context
Found during real-world upgrade from a pre-v1.107.0 installation. After upgrade, Caddy entered a restart loop because the Caddyfile still imported `Caddyfile.custom`. Separately, `Vector.php` and `CanastaFooterIcon.php` were recreated despite having been intentionally removed.

Fixes #517

### DIFF
--- a/internal/canasta/canasta.go
+++ b/internal/canasta/canasta.go
@@ -78,7 +78,7 @@ func CopyInstallationTemplate(destPath string) error {
 }
 
 // UpdateInstallationTemplate re-applies the embedded template during upgrade.
-// User-editable files are only created if missing (no-clobber).
+// User-editable files are skipped entirely (they may have been intentionally deleted).
 // CLI-managed files (READMEs, etc.) are always updated to match the current CLI version.
 func UpdateInstallationTemplate(destPath string) error {
 	logging.Print("Updating installation template files\n")
@@ -86,8 +86,8 @@ func UpdateInstallationTemplate(destPath string) error {
 }
 
 // copyTemplate walks the embedded installation template and copies files to destPath.
-// If upgrading is true, CLI-managed files are force-updated while user-editable files
-// use no-clobber. If upgrading is false, all files use no-clobber.
+// If upgrading is true, CLI-managed files are force-updated and user-editable files
+// are skipped. If upgrading is false (create), all files use no-clobber.
 func copyTemplate(destPath string, upgrading bool) error {
 	return fs.WalkDir(installationTemplate, "installation-template", func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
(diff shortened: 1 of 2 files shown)


internal/canasta/canasta.go
```diff
@@ -78,16 +78,16 @@ func CopyInstallationTemplate(destPath string) error {
...
 func UpdateInstallationTemplate(destPath string) error {
 	logging.Print("Updating installation template files\n")
 	return copyTemplate(destPath, true)
 }
 
 // copyTemplate walks the embedded installation template and copies files to destPath.
-// If upgrading is true, CLI-managed files are force-updated while user-editable files
-// use no-clobber. If upgrading is false, all files use no-clobber.
+// If upgrading is true, CLI-managed files are force-updated and user-editable files
+// are skipped. If upgrading is false (create), all files use no-clobber.
```
yaronkoren, line 90: Shouldn't this also change to "are skipped"?

yaronkoren: The second "use no-clobber", that is.

cicalese: It does — see lines 89-90 in the diff. The comment changed from "use no-clobber" to "are skipped".

yaronkoren: The second "use no-clobber".

cicalese: Correct. No-clobber is correct for create.

yaronkoren: I'm not sure you understood what I meant, but... it hardly matters.

yaronkoren (approved):

internal/canasta/canasta.go
cicalese: Maybe not. Are you talking about the 'no-clobber' on line 90? Or the one on line 126? Both of them are create behavior, which is no-clobber. Other than that, I'm not sure what you meant.

PR #523, opened 2026-03-02
Add test for displayVersion() in cmd/version

### LINKED ISSUE #501: Add tests for displayVersion()
## Context

`displayVersion()` in `cmd/version/version.go` formats the CLI version string for display. It has no test coverage.

## What to test

- Capture stdout and verify the output contains the expected version string
- If the function isn't easily testable in isolation, consider refactoring to return a string instead of printing directly, then test the returned string

## Guidance

- Look at `cmd/version/version.go` to understand the current implementation
- Follow patterns used elsewhere in `cmd/` tests if any exist

Ref #241

Mohammed-Thaha:
Add tests for the `DisplayVersion()` function in `cmd/version/version.go`:

This PR adds test coverage for the `displayVersion()` logic in `cmd/version/version.go`, as requested in #501.

**What changed in `version.go`:**

Extracted the version formatting logic into a new `DisplayVersion()` function that returns a string.
Previously the formatting and printing happened together inside `RunE`, which made it
difficult to test without capturing stdout. Now `RunE` simply calls `fmt.Println(DisplayVersion())`.
This is a minimal, non-breaking refactor the CLI output remains identical.

**Tests added in `version_test.go`:**

- **release_build** — sets Version to `v1.52.0` along with a commit hash and build time,
  then checks that the returned string contains all three values in the expected format.

- **dev_build_with_empty_version** — simulates a dev build by leaving Version empty,
  verifies the output falls back to `dev` and still includes the commit hash and build time.

- **all_fields_empty** — covers the edge case where Version, commit hash, and build time
  are all empty. Confirms the output defaults to `dev`.

- **only_commit_set** — sets only the commit hash with everything else empty,
  verifies the commit appears in the output alongside the `dev` version label.

Each test uses a `setVersionVars` helper that saves the original package-level variables
and restores them via `defer`, keeping tests isolated from each other.

Fixes #501


cmd/version/version.go
```diff
@@ -12,6 +12,13 @@ var (
 	sha1      string
 	buildTime string
 )
+func DisplayVersion() string {
+			v := Version
+			if v == "" {
+				v = "dev"
+			}
+			return fmt.Sprintf("Canasta CLI %s (commit %s, built %s)", v, sha1, buildTime)
+}
```
midnightCoder345: formatting does not make sense! Are you sure you wrote it?

cicalese:
Thanks for your contribution! The test logic is solid — good table-driven approach, proper save/restore of package-level variables, and nice coverage of edge cases.

There are a few style issues I'd like you to address before we merge:

1. **Indentation in `version.go`**: The `DisplayVersion()` function body is over-indented (tabs+spaces). It should use single-tab indentation to match the surrounding code.

2. **Missing blank line**: There should be a blank line before the `DisplayVersion()` function — it's currently right after the `var` block closing paren.

3. **Exported function unnecessary**: `DisplayVersion` is exported (capital D) but is only used within the `version` package. Please make it unexported (`displayVersion`) — the test file is in `package version` so it still has access.

4. **Missing newline at end of file**: The test file needs a trailing newline.

5. **Test file uses spaces instead of tabs**: The test file is indented with 4 spaces, but the rest of the codebase uses tabs. Please update to match.

Could you update the PR to fix these? Thanks!

Mohammed-Thaha:
Thanks for the feedback,
I have applied the requested changes.

cicalese (approved):
Thanks for your contribution!

PR #533, opened 2026-03-03
Fix install.sh directory collision on case-insensitive filesystems

### LINKED ISSUE #532: install.sh fails when a directory named Canasta exists in cwd
On case-insensitive filesystems (macOS), `install.sh` downloads to a file named `canasta` in the current working directory. If a directory named `Canasta/` exists (e.g., the Canasta repo), the download fails with `canasta: Is a directory`.

This also affects `canasta upgrade`, which downloads and runs `install.sh` via the self-update mechanism.

cicalese:
## Summary

- Download to a temp file (`mktemp`) instead of a bare `canasta` file in the current directory
- Fixes failure on macOS when a `Canasta/` directory exists in the working directory (case-insensitive collision)
- Also simplifies error handling by using a `trap` for cleanup and removing the interactive "keep file?" prompt

Closes #532

## Test plan

- [x] Run `canasta upgrade` from a directory containing a `Canasta/` subdirectory on macOS
- [x] Run `install.sh -v <version>` fresh install on Linux
- [x] Run `install.sh -v <version> -t /custom/path` to verify custom target still works

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -177,8 +177,14 @@ download_and_install() {
   local canasta_url=""
   local platform=""
   local binary_name=""
+  local tmpfile=""
 
   check_wget_show_progress
+
+  # Download to a temp file to avoid collisions with directories named "canasta"
+  # (e.g., on case-insensitive filesystems where Canasta/ exists).
+  tmpfile=$(mktemp "${TMPDIR:-/tmp}/canasta-download.XXXXXX")
+  trap 'rm -f "$tmpfile"' EXIT
 
   # Detect platform
   platform=$(detect_platform)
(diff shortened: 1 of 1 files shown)


yaronkoren:
Minor comment, but I think the comment in the code about avoiding name collisions won't really make sense, now that the downloaded file is being saved as "canasta-download.XXXXXX" and not "canasta".

cicalese:
Yeah, that was unnecessary historical context for the change that would be meaningless for future readers of the code. I fixed it.

yaronkoren (approved):
Great improvement!

PR #549, opened 2026-03-04
Added tests for SavePasswordToFile and UpdateEnvFile

### LINKED ISSUE #544: Add tests for SavePasswordToFile and UpdateEnvFile

Two functions in this package lack test coverage:

1. `SavePasswordToFile()` — write a password to a file in a directory, verify file contents and permissions
2. `UpdateEnvFile()` — merge a custom env file into the installation .env, verify correct precedence and that DB passwords and domain config are applied

Use `t.TempDir()` for filesystem isolation, following the existing test patterns in `canasta_test.go`.

Relates to #241

Deepan200201:
fixes https://github.com/CanastaWiki/Canasta-CLI/issues/544


yaronkoren:
@Deepan200201 - thank you for this PR, although please don't work on any more microtasks; we want to give other potential students a chance.

cicalese:
Thanks for your contribution. Some feedback on this PR:

1. **Table-driven tests**: The existing tests in this file use table-driven patterns with `t.Run` subtests. These should follow the same convention.
2. **`TestSavePasswordToFile`**: The permissions check (`&0600 != 0600`) will pass for virtually any file — it doesn't meaningfully verify anything. Also, only the happy path is covered; consider error cases (e.g. nonexistent directory).
3. **`TestUpdateEnvFile`**: The custom env file and the function arguments use the same password values, so the test doesn't actually verify that the function arguments take precedence. Using different values would make this a meaningful test.

Deepan200201:

cicalese (changes requested):
Thanks for addressing the earlier feedback — table-driven tests, error case for nonexistent directory, and different password values to verify argument priority all look good.

Two things to fix:

1. Several setup calls in the `TestUpdateEnvFile` loop silently discard errors:

```go
os.MkdirAll(subConfig, 0755)
os.WriteFile(filepath.Join(subConfig, "wikis.yaml"), []byte(wikiYamlContent), 0644)
os.WriteFile(baseEnvPath, []byte(tt.baseEnv), 0644)
os.WriteFile(customPath, []byte(tt.customEnv), 0644)
```

The outer setup checks these with `t.Fatal`. These should too — if setup fails silently, the test could pass or fail for the wrong reason.

2. This branch has merge conflicts with `main` and needs to be rebased.

cicalese (approved):
Review feedback has been addressed — table-driven tests, meaningful permission check, and the password precedence test now uses distinct values to verify argument priority. Looks good. Thank you.

Issue #542, opened 2026-03-04
CI: add golangci-lint workflow for PR validation


cicalese:
Add a CI workflow that runs golangci-lint on pull requests to catch code quality issues before merge.

Should coordinate with the existing `make lint` / `make prepare-lint` setup in the Makefile so that local and CI linting use the same configuration.

Relates to #241

cicalese:
### Lint violations by category (as of 2026-03-04, main branch)

Ran against 15 widely-used linters. Total: ~338 issues across 10 categories (govet, ineffassign, unused, gosimple, and unconvert report zero issues and can be enabled immediately).

| Linter | Count | Description | Example |
|--------|------:|-------------|---------|
| revive | 138 | General code quality (unused params, naming) | `parameter 'cmd' seems to be unused` |
| stylecheck | 54 | Go style conventions | `error strings should not be capitalized` |
| gosec | 53 | Security issues (file permissions, etc.) | `Expect WriteFile permissions to be 0600 or less` |
| errcheck | 21 | Unchecked error returns | `Error return value of os.WriteFile is not checked` |
| gocritic | 21 | Code improvements | `put a space between // and comment text` |
| staticcheck | 18 | Advanced static analysis | `possible nil pointer dereference` |
| unparam | 12 | Unused function parameters | `name always receives "test-wiki"` |
| gofmt | 9 | Formatting | `File is not properly formatted` |
| goimports | 9 | Import formatting/grouping | `File is not properly formatted` |
| misspell | 3 | Typos in comments/strings | `Contianers is a misspelling of Containers` |

### Cleanup checklist

Address each category then enable it in CI. Ordered by count (easiest first).

- [x] **misspell** (3) — #551
- [x] **gofmt** (9) — #551
- [x] **goimports** (9) — #551
- [x] **unparam** (12) — #553
- [x] **staticcheck** (18) — #555
- [x] **gocritic** (21) — #559
- [x] **errcheck** (21) — #557
- [x] **stylecheck** (54) — #561
- [x] **gosec** (53) — #563
- [x] **revive** (138) — #565

All violations addressed\! Next step: enable all 15 linters in CI via #540.

cicalese:
All 10 lint categories addressed (PRs #551–#567). CI enablement tracked in #540.

PR #557, opened 2026-03-04
Lint cleanup: fix all errcheck violations

### LINKED ISSUE #556: Lint cleanup: errcheck
Fix all `errcheck` violations reported by golangci-lint.

All violations are unchecked error returns from `os.WriteFile` and `os.MkdirAll` in test files, plus one `execute.Run` in production code (best-effort cleanup).

Part of #542.

cicalese:
Fix all `errcheck` violations reported by golangci-lint.

**Test files** (5 files): Wrap unchecked `os.WriteFile` and `os.MkdirAll` calls with `if err := ...; err != nil { t.Fatal(err) }`. In `devmode_test.go`, the setup function signatures were updated to accept `*testing.T` to enable error checking.

**Production code** (1 file): Mark one intentionally ignored `execute.Run` return (best-effort Docker container cleanup) with blank identifiers `_, _ =`.

Fixes #556
Part of #542

### DIFF
--- a/cmd/config/config_test.go
+++ b/cmd/config/config_test.go
@@ -163,7 +163,9 @@ func TestUpdateWikisYamlPorts(t *testing.T) {
 	// Create a temp dir with a wikis.yaml and .env
 	tmpDir := t.TempDir()
 	configDir := filepath.Join(tmpDir, "config")
-	os.MkdirAll(configDir, 0755)
+	if err := os.MkdirAll(configDir, 0755); err != nil {
+		t.Fatal(err)
+	}
 
 	wikisYaml := `wikis:
 - id: main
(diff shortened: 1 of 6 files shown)


yaronkoren:
Naive question, but don't all these Fatal() calls need a return call after them? Wasn't that the main change in the previous commit?

cicalese:
From Claude:

Good question! The `return` statements added in #555 (staticcheck) were specifically needed to suppress staticcheck's nil-pointer-dereference analysis. In that PR, the pattern was:

```go
wikiFlag := cmd.Flags().Lookup("wiki")
if wikiFlag == nil {
    t.Fatal("expected flag 'wiki'")
    return  // without this, staticcheck warns about the next line
}
if wikiFlag.Shorthand != "w" {  // staticcheck thinks wikiFlag could be nil here
```

Staticcheck doesn't model `runtime.Goexit()` (which `t.Fatal()` calls internally to stop the goroutine), so without the explicit `return`, it sees a code path where `wikiFlag` is nil and gets dereferenced on the next line.

In this PR, the pattern is different:

```go
if err := os.WriteFile(...); err != nil {
    t.Fatal(err)
}
```

Here `err` is scoped to the `if` block and nothing afterward depends on it, so there's no nil-dereference warning from any linter. No `return` is needed because there's no subsequent code that the linter thinks could use an invalid value.

yaronkoren:
Don't really understand this, but I'm glad there's a reason!

yaronkoren (approved):

PR #563, opened 2026-03-04
Lint cleanup: fix all gosec violations

### LINKED ISSUE #562: Lint cleanup: gosec
Fix all `gosec` violations reported by golangci-lint.

Violations fall into four categories:

- **G306** (file permissions): Change `0644` to `0600` in test `os.WriteFile` calls
- **G101** (hardcoded credentials): Suppress false positives on password character filter constant and VCL test strings
- **G115** (integer overflow): Add bounds check before `int` → `uint64` conversion in `gbToByte`
- **G204** (subprocess tainted input): Suppress false positives on `exec.Command` with compose binary path from system lookup

Part of #542.

cicalese:
Fix all `gosec` violations reported by golangci-lint.

**G306 — file permissions**: Excluded globally in `.golangci.yml` since `0644` is the project's intentional default file permission. No per-line `//nolint` directives needed.

**G101 — hardcoded credentials** (3 false positives): Suppress with `//nolint:gosec` on own line:
- `problematicPasswordChars` constant — character filter list, not credentials
- `vclWithoutBypass`/`vclWithBypass` — VCL config strings in tests

**G115 — integer overflow** (1): Add `gb < 0` guard before `int` → `uint64` conversion in `gbToByte`.

**G204 — subprocess tainted input** (6): Suppress with `//nolint:gosec` on own line for `exec.Command` calls using `compose.Path` (resolved from system binary lookup via `exec.LookPath`).

Fixes #562
Part of #542

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,4 @@
+linters-settings:
+  gosec:
+    excludes:
+      - G306
--- a/cmd/backup/backup_test.go
+++ b/cmd/backup/backup_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/CanastaWiki/Canasta-CLI/internal/canasta"
 	"github.com/CanastaWiki/Canasta-CLI/internal/farmsettings"
+	"github.com/CanastaWiki/Canasta-CLI/internal/permissions"
 )
 
 func writeWikisYaml(t *testing.T, dir string, wikis []farmsettings.Wiki) {
@@ -22,7 +23,8 @@ func writeWikisYaml(t *testing.T, dir string, wikis []farmsettings.Wiki) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if err := os.WriteFile(filepath.Join(configDir, "wikis.yaml"), data, 0600); err != nil {
+	//nolint:gosec
+	if err := os.WriteFile(filepath.Join(configDir, "wikis.yaml"), data, permissions.FilePermission); err != nil {
 		t.Fatal(err)
 	}
 }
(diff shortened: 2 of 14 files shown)


yaronkoren:
I don't know what this gosec checker is, but wouldn't it be better to just remove it (if that's possible), if seemingly everything violates it?

cicalese:
Claude's analysis:

gosec is a Go security linter — it catches things like hardcoded credentials, unsafe file permissions, potential integer overflows, and command injection via `exec.Command`. Most of the violations here were G306 (file permissions in test files), which are low-risk false positives. But the other categories it caught are genuinely useful:

- **G115** — integer overflow in `gbToByte` (missing negative check before `int` → `uint64` conversion)
- **G204** — flagging `exec.Command` with variable arguments (we suppress these because the inputs come from `exec.LookPath`, but it's good to have them explicitly acknowledged)
- **G101** — potential hardcoded credentials (false positives here, but valuable in production code)

Once these are addressed, gosec runs clean — it won't flag new code unless there's a genuine concern. The `//nolint:gosec` directives document our intentional exceptions.

yaronkoren:
Okay, thanks, Claude. :) But seriously, this just seems to make the code more complex (presumably many added lines in the future will require their own "nolint" comment as well), for just minimal payoff - I'm not sure that having gbToByte() return 0 for a negative input is even an improvement

cicalese:
It turns out we can exclude the most annoying warnings globally. Working on it.

yaronkoren (approved):
Thanks for adding that exclusion! This looks reasonable now.

Issue #568, opened 2026-03-05
Code quality audit (Round 6): dead code, bugs, duplication, and architecture


cicalese:
A critical analysis of the Canasta-CLI codebase identified the following issues. Items are grouped by category with checkboxes for tracking.

## Dead Code / Duplication

- [x] **Duplicate `getWikiIDs`**: `cmd/backup/backup.go:91-98` reimplements `farmsettings.GetWikiIDs`. Replace with a call to the existing function. (#569)
- [x] **Duplicate `shellEscape`**: `cmd/upgrade/mysql_to_mariadb.go:193-195` is identical to `orchestrators.ShellQuote`. Replace with a call to the existing function. (#569)
- [x] **`io.MultiWriter` with single writer**: `internal/execute/execute.go:49-50` wraps a single writer in `MultiWriter` — leftover from when multiple writers were used. Use the writer directly. (#569)
- [x] **Compose `Path` branching boilerplate**: `internal/orchestrators/compose.go` has ~15 methods that each check `compose.Path != ""` and branch into two nearly identical `exec.Command` calls. Consolidate into a single `runCompose(installPath, args...)` helper. (#573)
- [x] **Stack file walking logic duplicated 4×**: `walkStackFiles` and `UpdateStackFiles` in both `compose.go` and `kubernetes.go` are near-identical. Extract a shared utility. (#574)
- [x] **Restart logic duplicated in config set/unset**: `cmd/config/set.go:181-204` and `cmd/config/unset.go:79-96` have identical restart sequences (UpdateConfig → Stop → optional recreateKindCluster → Start). Extract a shared helper. (#570)
- [x] **`resolveKey` in wrong file**: Defined in `cmd/config/get.go:63-74` but used across `set.go`, `unset.go`, and `get.go`. Move to a shared file (e.g., `helpers.go`). (#569)

## Bugs

- [x] **`GetEnvVariable` doesn't skip comments**: `internal/canasta/canasta.go:784-806` parses comment lines (e.g., `# FOO=bar`) as key-value pairs. Add comment and blank-line filtering. (#570)
- [x] **`SaveEnvVariable` can produce duplicate entries**: `internal/canasta/canasta.go:757-781` appends new keys without checking if they already exist elsewhere in the file. (#572)
[...]

PR #569, opened 2026-03-05
Fix dead code, duplication, and bugs from code quality audit

cicalese:
## Summary

Addresses the low-hanging fruit from #568 (code quality audit Round 6):

- **Remove duplicate `getWikiIDs`**: `cmd/backup/backup.go` reimplemented `farmsettings.GetWikiIDs` — replaced with a reference to the existing function
- **Remove duplicate `shellEscape`**: `cmd/upgrade/mysql_to_mariadb.go` duplicated `orchestrators.ShellQuote` — replaced with a call to the existing function
- **Remove unnecessary `io.MultiWriter` wrapper**: `internal/execute/execute.go` wrapped a single writer in `MultiWriter` — use the writer directly
- **Move shared helpers to dedicated file**: `resolveKey` (from `get.go`) and `isKnownKey` (from `set.go`) moved to `cmd/config/helpers.go` since both are used across `get.go`, `set.go`, and `unset.go`
- **Fix `SetupDevEnvironment` duplicate `.env` entries**: `internal/devmode/devmode.go` appended `DEV_CODE_PATH` and `CANASTA_IMAGE` via raw byte append, creating duplicates on re-enable — now uses `canasta.SaveEnvVariable`
- **Read `.env` once in `generateKustomization`**: `internal/orchestrators/kubernetes.go` called `canasta.GetEnvVariable` three times for observability, elasticsearch, and image override checks — consolidated into a single read
- **Return error from `getRepoURL` on empty env**: `cmd/backup/backup.go` produced `s3:/` when no backup repository was configured — now returns a clear error message

## Test plan

- [x] All existing tests pass (`go test ./...`)
- [x] All 15 golangci-lint linters pass clean
- [x] Backup tests updated for new `getRepoURL` error return (empty env cases now expect errors)

### DIFF
--- a/cmd/backup/backup.go
+++ b/cmd/backup/backup.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/CanastaWiki/Canasta-CLI/internal/canasta"
 	"github.com/CanastaWiki/Canasta-CLI/internal/config"
-	"github.com/CanastaWiki/Canasta-CLI/internal/farmsettings"
 	"github.com/CanastaWiki/Canasta-CLI/internal/logging"
 	"github.com/CanastaWiki/Canasta-CLI/internal/orchestrators"
 )
@@ -46,7 +45,10 @@ and RESTIC_PASSWORD to be configured in the installation's .env file.`,
 			if envErr != nil {
 				return envErr
 			}
-			repoURL = getRepoURL(envVariables)
+			repoURL, err = getRepoURL(envVariables)
+			if err != nil {
+				return err
+			}
 
 			orch, err = orchestrators.New(instance.Orchestrator)
 			if err != nil {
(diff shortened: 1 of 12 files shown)


cmd/backup/backup.go
```diff
@@ -88,14 +96,7 @@ func runBackup(orch orchestrators.Orchestrator, installPath, envPath string, vol
...
 // getWikiIDs reads wikis.yaml and returns all wiki IDs.
-func getWikiIDs(installPath string) ([]string, error) {
-	yamlPath := filepath.Join(installPath, "config", "wikis.yaml")
-	ids, _, _, err := farmsettings.ReadWikisYaml(yamlPath)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read wikis.yaml: %w", err)
-	}
-	return ids, nil
-}
+var getWikiIDs = farmsettings.GetWikiIDs
```
yaronkoren: What's the point of creating this "alias", if that's what it is? Why not just call that function directly?

cicalese: Actually, no point. Good catch. It's fixed now.

cicalese: But, now we need to move some tests to farmsettings.

cicalese: OK, fixed. Ready for review.

yaronkoren (approved):

PR #570, opened 2026-03-05
Fix nil panic, duplicate restart logic, and .env comment parsing

cicalese:
## Summary

Continuation of #568 (code quality audit), stacked on #569.

- **Nil panic in `UpdateEnvFile`**: `canasta.go:183` accessed `domainNames[0]` without checking if the slice was empty. Added defensive check.
- **Duplicate restart logic in config set/unset**: Identical restart sequences (UpdateConfig → Stop → optional kind cluster recreation → Start) in `set.go` and `unset.go` extracted into shared `restartInstance` helper in `helpers.go`.
- **`GetEnvVariable` doesn't skip comments**: Lines like `# FOO=bar` were parsed as key-value pairs. Now skips comment lines (starting with `#`) and blank lines.

## Test plan

- [x] All existing tests pass (`go test ./...`)
- [x] All 15 golangci-lint linters pass clean
- [x] Added test assertion that comment lines with `=` signs are excluded from parsed env vars

### DIFF
--- a/cmd/config/helpers.go
+++ b/cmd/config/helpers.go
@@ -1,6 +1,12 @@
 package config
 
-import "strings"
+import (
+	"fmt"
+	"strings"
+
+	"github.com/CanastaWiki/Canasta-CLI/internal/config"
+	"github.com/CanastaWiki/Canasta-CLI/internal/orchestrators"
+)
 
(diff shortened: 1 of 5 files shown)


yaronkoren (approved):
Another great simplification!

PR #573, opened 2026-03-05
Extract runCompose and composeCommand helpers

cicalese:
## Summary
- Every `ComposeOrchestrator` method duplicated the same `if compose.Path != ""` branch to handle standalone compose binary vs `docker compose` subcommand
- Extracted two helpers:
  - `runCompose` — for methods using `execute.Run` (8 call sites)
  - `composeCommand` — for methods using `exec.Command` (7 call sites)
- Net -95 lines from a single file, purely mechanical refactoring

Fixes an item from #568.


yaronkoren (approved):
Wow, 95 lines removed!

PR #574, opened 2026-03-05
Extract shared stack file walking logic

cicalese:
## Summary
- `walkStackFiles` and `UpdateStackFiles` were near-identical (97-98%) in both `compose.go` and `kubernetes.go`, differing only in the embedded FS and root path
- Extracted `writeStackFiles` and `updateStackFiles` into a new `stackfiles.go` and made both orchestrators delegate to them
- Net -61 lines (removed 159, added 98 in shared file)

Fixes an item from #568.

### DIFF
--- a/internal/orchestrators/compose.go
+++ b/internal/orchestrators/compose.go
@@ -1,9 +1,7 @@
 package orchestrators
 
 import (
-	"bytes"
 	"fmt"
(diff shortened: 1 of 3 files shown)


yaronkoren (approved):
Looks like a good refactoring!

PR #576, opened 2026-03-05
Use %w instead of %s/%v for error wrapping

cicalese:
## Summary
- 13 `fmt.Errorf` calls wrapped errors with `%s` or `%v`, which discards the error chain and prevents `errors.Is`/`errors.As` from working
- Switch all to `%w` for proper error wrapping per Go conventions
- 9 files changed, purely mechanical substitution

Fixes an item from #568.


yaronkoren (approved):
Great improvement!

PR #577, opened 2026-03-05
Standardize path resolution at command entry points

cicalese:
## Summary

- Moves relative-to-absolute path resolution from 5 internal functions to the cmd/ layer where the working directory is already known
- Removes the `workingDir` parameter from `UpdateEnvFile`, `CopyWikiSettingFile`, `CopyGlobalSettingFile`, `CopyComposerFile`, and `CopyOverrideFile`
- Removes the `WorkingDir` field from `createOptions` and `AddWikiOptions` structs
- Each command resolves all file paths once at the top of `RunE` using a compact loop

## Details

Previously, relative path resolution was scattered across 5 internal functions, each independently checking `filepath.IsAbs` and joining with a `workingDir` parameter passed down from the command layer. This was inconsistent (some commands resolved paths early, others deep inside called functions) and required threading `workingDir` through option structs.

Now all file paths are resolved to absolute at the command entry point (`RunE`), and internal functions always receive absolute paths.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./...` passes
- [x] `golangci-lint run ./...` passes

Ref #568

### DIFF
--- a/cmd/add/add.go
+++ b/cmd/add/add.go
@@ -131,6 +131,13 @@ an existing database dump instead of running the installer.`,
 				fmt.Printf("Generated admin password for wiki '%s'\n", wikiID)
 			}
 
+			// Resolve relative file paths to absolute
+			for _, p := range []*string{&databasePath, &wikiSettingsPath} {
+				if *p != "" && !filepath.IsAbs(*p) {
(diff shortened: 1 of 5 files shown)


yaronkoren (approved):
Makes sense - make the paths absolute from the beginning!

PR #582, opened 2026-03-05
Replace shell-based self-update with pure Go (#581)

### LINKED ISSUE #581: Self-update fails with Permission denied for non-root users
## Problem

`canasta upgrade` fails during CLI self-update for non-root users (e.g. `mediawiki` service accounts) with a confusing error:

```
Updating CLI from v3.3.2 to v3.4.0...
-----
Checking the version...
Version 3.4.0 has already been specified; proceeding with installation.
Detected platform: linux-amd64
Downloading Canasta CLI version 3.4.0 for linux-amd64...
/tmp/canasta-download.FGR3Nn: Permission denied
Platform-specific binary not found. Trying generic binary (backward compatibility)...
/tmp/canasta-download.FGR3Nn: Permission denied
Download failed. The version you specified might not exist.
Please use '-l' or '--list' flag to see the available versions or try again.
CLI update failed: install script failed with status 1
```

The self-update calls `install.sh`, which uses `sudo wget` to download the binary to a temp file. This fails when:
- The user doesn't have `sudo` access
- `sudo` can't prompt for a password in a non-interactive context

The "Please use '-l' or '--list' flag" message is misleading in this context — the download failed due to permissions, not a missing version.

## Solution

Replace the shell-based self-update (`install.sh` + `bash` + `sudo wget`) with a pure Go implementation that downloads the binary directly via `net/http`, preserving permissions and doing an atomic rename. This eliminates dependencies on `bash`, `wget`, and `sudo`, and provides clear error messages when permissions are insufficient.

`install.sh` is kept for standalone first-time installation.

cicalese:
## Summary

- Replace `install.sh`/`bash`/`sudo wget` self-update with pure Go: download via `net/http`, then `os.Rename` or `sudo mv`
- Eliminate dependencies on `bash` and `wget` during self-update
- Use `sudo` only for the final move when the binary is in a privileged directory (e.g. `/usr/local/bin`)
- Remove `findBash()` and `installShURL` constant (no longer needed)

## Details

The self-update called `install.sh`, which used `sudo wget` to download the binary to a temp file. This failed for users without `sudo` access (e.g. `mediawiki` service accounts) with a misleading "Permission denied" / "Please use '--list' flag" error.

The new implementation:
1. Downloads the platform-specific binary via `net/http.Get` to the OS temp dir
2. Preserves permissions of the existing binary via `os.Stat`/`os.Chmod`
3. Tries `os.Rename` first (works if same filesystem and user has write access)
4. Falls back to `sudo mv` if rename fails, prompting for a password only when needed

This avoids running the entire upgrade as root (which could cause ownership problems on config files) while still supporting privileged install locations.

`install.sh` is kept for standalone first-time installation.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./...` passes
- [x] `golangci-lint run ./...` passes
- [x] Manual: build with a fake old version, run `canasta upgrade`, verify it downloads and replaces the binary
- [x] Manual: test as non-root user with binary in `/usr/local/bin` — verify sudo prompt appears only for the move

Fixes #581

### DIFF
--- a/internal/selfupdate/selfupdate.go
+++ b/internal/selfupdate/selfupdate.go
@@ -72,18 +72,9 @@ func downloadAndInstall(latestVersion, execPath string) error {
 		latestVersion, runtime.GOOS, runtime.GOARCH,
 	)
 
-	// Try to create temp file in the same directory as the target binary
-	// (same filesystem enables atomic os.Rename). Fall back to OS temp dir
-	// if the directory isn't writable — sudo mv will handle the final move.
-	dir := filepath.Dir(execPath)
-	tmpFile, err := os.CreateTemp(dir, ".canasta-update-*")
+	tmpFile, err := os.CreateTemp("", ".canasta-update-*")
 	if err != nil {
-		if os.IsPermission(err) {
-			tmpFile, err = os.CreateTemp("", ".canasta-update-*")
-		}
-		if err != nil {
-			return fmt.Errorf("failed to create temp file: %w", err)
-		}
+		return fmt.Errorf("failed to create temp file: %w", err)
 	}
 	tmpPath := tmpFile.Name()
 	defer os.Remove(tmpPath)
--- a/internal/selfupdate/selfupdate.go
+++ b/internal/selfupdate/selfupdate.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


yaronkoren (approved):
Pretty cool that Go has all these native commands!

PR #585, opened 2026-03-05
test: add getLatestVersion tests with httptest mock

### LINKED ISSUE #545: Add tests for selfupdate getLatestVersion
**File:** `internal/selfupdate/selfupdate.go`
**Test file:** `internal/selfupdate/selfupdate_test.go` (new)

`getLatestVersion()` queries the GitHub API for the latest release tag. This can be tested using `net/http/httptest` to serve a mock JSON response, then verifying the parsed tag name.

Test cases:
1. Valid release response — returns correct tag
2. Empty `tag_name` — returns error
3. Non-200 status code — returns error
4. Invalid JSON — returns error

The function currently uses a package-level `githubAPIURL` constant. It may need to be refactored to accept a URL parameter or use a package-level variable to make it testable with httptest.

Relates to #241

Mohammed-Thaha:
Fixes #545
Small refactor and new tests for getLatestVersion().

**What changed:**

githubAPIURL: const → var so tests can point it to a mock server
Added selfupdate_test.go with 4 tests using httptest

**Tests added:**

- Valid response → returns correct tag
- Empty tag_name → returns error
- Non-200 status → returns error
- Invalid JSON → returns error

### DIFF
--- a/internal/selfupdate/selfupdate.go
+++ b/internal/selfupdate/selfupdate.go
@@ -14,9 +14,7 @@ import (
 	"github.com/CanastaWiki/Canasta-CLI/cmd/version"
 )
 
-// githubAPIURL is a variable (not a constant) so that tests can override it
-// with a local httptest server URL.
-var githubAPIURL = "https://api.github.com/repos/CanastaWiki/Canasta-CLI/releases/latest"
+const githubAPIURL = "https://api.github.com/repos/CanastaWiki/Canasta-CLI/releases/latest"
 
 type githubRelease struct {
 	TagName string `json:"tag_name"`
@@ -134,7 +132,7 @@ func reexec(execPath string) error {
 }
 
 func getLatestVersion() (string, error) {
-	//nolint:gosec // URL is a package-level variable intentionally overridable in tests
+	//nolint:gosec // URL is a compile-time constant
 	resp, err := http.Get(githubAPIURL)
 	if err != nil {
 		return "", fmt.Errorf("failed to check for updates: %w\nPlease check your network connectivity", err)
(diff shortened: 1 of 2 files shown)


yaronkoren:
@Mohammed-Thaha - thank you for this PR! Please don't work on any more microtasks, though - we want to save the rest for other applicants.

Mohammed-Thaha:
> @Mohammed-Thaha - thank you for this PR! Please don't work on any more microtasks, though - we want to save the rest for other applicants.

Thank you for the clarification.This is my second and final beginner microtask. I won’t work on any more microtasks and will start working on my proposal

cicalese (changes requested):
Clean tests with good coverage of all code paths.

One concern: changing `githubAPIURL` from `const` to `var` weakens the production code just to make it testable. The URL should remain a constant.

Instead, extract the response parsing into a separate function and test that directly:

```go
func parseLatestVersion(body io.Reader) (string, error) {
	var release githubRelease
	if err := json.NewDecoder(body).Decode(&release); err != nil {
		return "", fmt.Errorf("failed to parse release info: %w", err)
	}
	if release.TagName == "" {
		return "", fmt.Errorf("no tag_name found in latest release")
	}
	return release.TagName, nil
}
```

Then `getLatestVersion` calls `parseLatestVersion(resp.Body)`, and tests can call `parseLatestVersion` directly with `strings.NewReader(...)` — no mock server needed, no global state mutation. The HTTP status check can stay in `getLatestVersion` since it's trivial standard library code.

Mohammed-Thaha:
> Clean tests with good coverage of all code paths.
> 
> One concern: changing `githubAPIURL` from `const` to `var` weakens the production code just to make it testable. The URL should remain a constant.
> 
> Instead, extract the response parsing into a separate function and test that directly:
> 
> ```go
> func parseLatestVersion(body io.Reader) (string, error) {
> 	var release githubRelease
> 	if err := json.NewDecoder(body).Decode(&release); err != nil {
> 		return "", fmt.Errorf("failed to parse release info: %w", err)
> 	}
> 	if release.TagName == "" {
> 		return "", fmt.Errorf("no tag_name found in latest release")
> 	}
> 	return release.TagName, nil
> }
> ```
> 
> Then `getLatestVersion` calls `parseLatestVersion(resp.Body)`, and tests can call `parseLatestVersion` directly with `strings.NewReader(...)` — no mock server needed, no global state mutation. The HTTP status check can stay in `getLatestVersion` since it's trivial standard library code.

@cicalese thanks for the review
Addressed your feedback. please take another look.

cicalese (changes requested):
Looks good — addressed all the feedback cleanly. One small thing: the `//nolint:gosec` comment on line 135 can be removed. It was only needed when `githubAPIURL` was a `var`; as a `const`, gosec doesn't flag it.

cicalese (approved):
LGTM! Thank you for your contribution!

PR #587, opened 2026-03-06
test: add missing unit tests for extensionsskins disable and CheckEnabled

### LINKED ISSUE #541: Add tests for extension/skin disable per-wiki and CheckEnabled() for skins
**File:** `internal/extensionsskins/extensionsskins.go` — `Disable()`, `CheckEnabled()`
**Test file:** `internal/extensionsskins/extensionsskins_test.go`

All existing disable tests use extension constants at the global level. Missing coverage:

1. Disabling a skin (via `skinConstants`) — verify skin is removed from config
2. Disabling a per-wiki extension — verify the per-wiki settings file is updated and global config is untouched
3. `CheckEnabled()` for a skin — verify it returns the correct skin name

No mocking needed — use `t.TempDir()` and real filesystem, following the existing test pattern.

Split from #495. Relates to #241

Mounil2005:
**Description:**

The Canasta CLI lets you enable and disable MediaWiki extensions and skins. While the underlying logic was working correctly, several scenarios had no automated tests covering them:

- Disabling a skin was never tested
- Disabling an extension scoped to a specific wiki was never tested
- Checking whether a skin is enabled was never tested
- Without these tests, a future code change could silently break any of these behaviors without anyone knowing until it hit production.

This PR adds tests for all three missing cases, and also fixes an existing test that used hardcoded Unix-style paths, which caused it to fail on Windows.

**Related issue(s):**

Fixes #541

**Notes for reviewer:**

All 17 tests passing. No changes to production code, test file only.

**Checklist:**
[x] Tested (go test ./internal/extensionsskins/... -v)

### DIFF
--- a/internal/extensionsskins/extensionsskins_test.go
+++ b/internal/extensionsskins/extensionsskins_test.go
@@ -330,7 +330,10 @@ func TestDisableSkin(t *testing.T) {
 		t.Fatalf("Disable skin: %v", err)
 	}
 
-	cfg, _ := readConfig(configPath(inst.Path, ""))
+	cfg, err := readConfig(configPath(inst.Path, ""))
+	if err != nil {
+		t.Fatalf("readConfig: %v", err)
+	}
 	if slices.Contains(cfg.Skins, "Vector") {
 		t.Error("Vector should be removed from skins")
 	}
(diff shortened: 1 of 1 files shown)


cicalese (changes requested):
Good tests that fill real gaps — skin disable, per-wiki extension disable, and skin `CheckEnabled` were all untested. The Windows path fix with `filepath.Join` is also a good catch.

One issue: `readConfig` errors are silently discarded in both `TestDisableSkin` and `TestDisablePerWikiExtension`:

```go
cfg, _ := readConfig(configPath(inst.Path, ""))
```

```go
wikiCfg, _ := readConfig(configPath(inst.Path, "docs"))
```

If `readConfig` fails, the test would pass vacuously or panic on nil. These should check the error with `t.Fatal`.

Mounil2005:
Good catch! Updated TestDisableSkin and TestDisablePerWikiExtension to properly handle the error returned by readConfig instead of discarding it. If the config file can't be read, the test now fails immediately with a descriptive message rather than silently continuing with an empty struct.

cicalese (approved):

PR #598, opened 2026-03-09
Remove legacy .gitignore from config/ and config/settings/

### LINKED ISSUE #597: gitops init does not track config/settings due to legacy .gitignore
Old Canasta-DockerCompose installations have `.gitignore` files in `config/` and `config/settings/` that ignore everything except `.gitignore` itself. When `canasta gitops init` runs, these files prevent `git add` from tracking PHP settings and other configuration files under `config/settings/`.

PR #595 added removal of legacy `.gitignore` files from `extensions/` and `skins/`, but missed `config/` and `config/settings/`.

cicalese:
## Summary

- Extends `removeLegacyGitignores` to also clean `config/` and `config/settings/`, not just `extensions/` and `skins/`
- Without this, old Canasta-DockerCompose installations have `.gitignore` files that prevent `gitops init` from tracking PHP settings and configuration files

## Related issue(s)

Fixes #597

## Test plan

- [x] Confirmed that `config/.gitignore` and `config/settings/.gitignore` exist on a production Canasta instance upgraded from the old template
- [x] Run `canasta gitops init` on an installation with these legacy files — both removed with informational messages and `config/settings/` contents are fully tracked

### DIFF
--- a/cmd/gitops/init.go
+++ b/cmd/gitops/init.go
@@ -330,12 +330,18 @@ func convertToSubmodules(installPath, dirName string) error {
 	return nil
 }
 
-// removeLegacyGitignores removes .gitignore files from extensions/ and skins/
-// that were created by the old Canasta-DockerCompose installation template.
-// These files ignore everything except .gitignore itself, which would prevent
-// gitops from tracking extensions and skins.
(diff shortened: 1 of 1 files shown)


yaronkoren (approved):
Looks good!

PR #627, opened 2026-03-10
Add git-crypt and git config to CI integration workflow

cicalese:
## Summary

- Installs `git-crypt` in the integration workflow (required by gitops test)
- Configures `git user.name`/`user.email` in CI (required for git commit in gitops init)
- Adds `t.Skip` guard to gitops test when `git-crypt` is unavailable

Note: the underlying code issues are fixed separately in #630 (git config check) and #631 (backup restore overwrite). This PR ensures the CI environment has the necessary tools and config.

## Test plan

- [ ] Integration workflow runs without git-crypt/git-config errors
- [ ] All integration tests pass

Fixes the failure in https://github.com/CanastaWiki/Canasta-CLI/actions/runs/22927260846

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -15,6 +15,8 @@ jobs:
       uses: actions/setup-go@v5
       with:
         go-version: '1.25'
+    - name: Install git-crypt
+      run: sudo apt-get update && sudo apt-get install -y git-crypt
     - name: Add runner to www-data group
       # Canasta containers run as www-data, so bind-mounted directories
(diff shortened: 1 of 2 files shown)


yaronkoren:
Not that I really need to understand this, but - if git-crypt is being installed, why does the test need to check whether it's installed?

cicalese:
In case the install fails, although in that case, the test really should fail. I'm going to remove that check.

cicalese:
Done.

yaronkoren (approved):

PR #636, opened 2026-03-11
Fix gitops status truncating modified file paths

### LINKED ISSUE #635: gitops status truncates first character of modified file paths
## Description

`canasta gitops status` truncates the first character of file paths for modified (but not untracked) files.

For example:
```
Uncommitted changes (6 files):
  onfig/.smw.json                    ← should be config/.smw.json
  xtensions/wikivisor/SkinSettings.php    ← should be extensions/...
```

Untracked files (`??` status) display correctly.

## Root Cause

`HasUncommittedChanges` in `internal/gitops/git.go` calls `strings.TrimSpace()` on each `git status --porcelain` line before extracting the filename at `line[3:]`. For unstaged modifications, the porcelain format uses a space as the first status character (e.g., ` M config/.smw.json`). Trimming converts this to `M config/.smw.json` — only 2 prefix characters instead of 3 — so `line[3:]` clips the first character of the path.

## Fix

Remove the `TrimSpace` on individual lines. The raw porcelain format always has exactly 3 prefix characters (`XY `) before the filename.

cicalese:
## Summary

- Fix `HasUncommittedChanges` stripping the first character from modified file paths in `canasta gitops status` output (e.g. `onfig/.smw.json` instead of `config/.smw.json`)
- Two bugs: per-line `TrimSpace` removed the leading space from ` M` status codes, and whole-output `TrimSpace` also stripped it from the first line
- Replace `TrimSpace` with `TrimRight` for the whole output, remove per-line trimming entirely
- Add unit tests covering porcelain line parsing and leading-space preservation

Fixes #635

## Test plan

- [x] Run `canasta gitops status` on an instance with both unstaged modifications and untracked files
- [x] Verify all file paths are displayed correctly with no truncation
- [x] `go test ./internal/gitops/` passes
- [x] Local integration test with registered Canasta instance confirms ` M config/.smw.json` displays as `config/.smw.json`

### DIFF
--- a/internal/gitops/git.go
+++ b/internal/gitops/git.go
@@ -154,7 +154,9 @@ func HasUncommittedChanges(path string) (bool, []string, error) {
 	if err != nil {
 		return false, nil, fmt.Errorf("git status: %w", err)
 	}
-	output = strings.TrimSpace(output)
+	// Trim only trailing whitespace — leading spaces are significant
+	// in porcelain format (e.g. " M file" means unstaged modification).
+	output = strings.TrimRight(output, " \t\n\r")
 	if output == "" {
 		return false, nil, nil
 	}
--- a/internal/gitops/git.go
+++ b/internal/gitops/git.go
@@ -160,16 +160,14 @@ func HasUncommittedChanges(path string) (bool, []string, error) {
 	}
 	var files []string
 	for _, line := range strings.Split(output, "\n") {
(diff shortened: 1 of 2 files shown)


yaronkoren (approved):
I never knew the term "porcelain" for Git - though apparently I've been using porcelain this whole time!

Issue #608, opened 2026-03-10
Test: add unit tests for repairBrokenSubmodules


cicalese:
## Context

`repairBrokenSubmodules()` in `cmd/gitops/init.go` reads `.gitmodules` and re-registers any submodules that were committed as regular trees (040000) instead of gitlinks (160000 commit). It has no unit test coverage.

## What to test

Add tests in `cmd/gitops/init_test.go` covering:

1. **Broken submodule repaired** — create a git repo with a `.gitmodules` entry and the extension committed as a regular tree (not a submodule). Verify it gets re-registered as a proper submodule (160000 commit in `git ls-tree`).
2. **Healthy submodule unchanged** — create a repo with a properly registered submodule. Verify it returns 0 repaired.
3. **No .gitmodules file** — verify 0 repaired and no error.
4. **Empty .gitmodules** — verify 0 repaired and no error.
5. **Broken gitlink file removed** — create a `.git` file (gitlink) in the extension dir. Verify it is cleaned up during repair.

## Hints

- Use `t.TempDir()` for test directories
- To create the broken state: `git init`, add files, write `.gitmodules` manually, `git add -A`, `git commit` — this commits extensions as trees
- To verify repair: run `git ls-tree HEAD extensions/` and check for `160000` prefix
- The function signature is `repairBrokenSubmodules(installPath string) (int, error)`
- These tests need network access to clone from the submodule URL in `.gitmodules`

Part of #241

ChefYeshpal:
hello, I have tried to solve this issue (test) in my own fork ([here](https://github.com/ChefYeshpal/Canasta-CLI)).
If possible could you please look at it and tell me if something is missing? or if I've misunderstood something?
[Here is the latest commit by me (as of 11 March 2026)](https://github.com/ChefYeshpal/Canasta-CLI/commit/6cd795b9f6cd2d625bf730c24e73aee194f2d96a)

cicalese:
Thanks for taking this on! The test coverage looks solid — you've covered all five scenarios from the issue description, and the helper functions are well-structured.

A few suggestions before opening a PR:

[...]

PR #639, opened 2026-03-12
test(gitops): add unit tests for repairBrokenSubmodules

ChefYeshpal:
## Summary
Adds unit test coverage for repairBrokenSubmodules in cmd/gitops/init_test.go.

## Whats done
- broken submodule committed as a tree is repaired
- healthy submodule is unchanged
- missing .gitmodules returns 0 with no error
- empty .gitmodules returns 0 with no error
- broken gitlink file case is repaired correctly

## Notes
- verifies repaired submodules via git index mode (160000)
- moves git test helpers into cmd/gitops/testutil_test.go for reuse

Reference: #608

### DIFF
--- a/cmd/gitops/init_test.go
+++ b/cmd/gitops/init_test.go
@@ -1,7 +1,10 @@
 package gitops
 
 import (
+	"bytes"
+	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


ChefYeshpal:
I didn't check via `golangci-lint` as I didn't have that installed on my system (yet), should I still code correct it and do another pull request?

cicalese:
Thanks for addressing the review feedback — the changes look good. All three suggestions from the issue discussion have been incorporated:

1. The extra `git add -A` / `git commit` is gone — you're now correctly checking the index directly with `ls-files --stage` after repair, which is cleaner.

2. The "broken gitlink file removed" test now asserts 160000 in the index rather than inspecting the `.git` file content — verifying the actual end result.

3. The helpers are in `testutil_test.go` for reuse by future tests.

CI passes (lint + tests). Looks good to merge.

cicalese (approved):
LGTM. Thank you for your contribution!

ChefYeshpal:
nws :> (I dont really know how to close issues haha)

PR #643, opened 2026-03-13
Add wikis.yaml templating for gitops

### LINKED ISSUE #642: gitops: wikis.yaml URLs are host-specific but tracked as shared config
## Problem

`wikis.yaml` is tracked in the gitops repo as shared configuration, but the `url` field in each wiki entry is host-specific. For example, a production server uses `wiki.example.com` while a local dev instance uses `localhost`.

This means a developer who restores a backup, joins the gitops repo, and changes the URLs in `wikis.yaml` to `localhost` for local testing will push those localhost URLs to the shared repo if they ever stage `config/wikis.yaml`, overwriting the production URLs.

With the explicit `gitops add` command (#637), this is mitigated since files must be explicitly staged. But it's still a footgun — one careless `gitops add config/` overwrites prod URLs. And `gitops status` will always show `wikis.yaml` as modified on any host with different URLs, creating noise.

## Proposed Solution

Apply the same pattern used for `.env` → `env.template` + per-host `vars.yaml`:

1. Create `wikis.yaml.template` with `{{placeholders}}` for URLs
2. Store actual URLs in the per-host `hosts/{name}/vars.yaml`
3. Render `config/wikis.yaml` from the template on `gitops pull`
4. Add `config/wikis.yaml` to `.gitignore` (like `.env`)
5. Update `gitops init` and `gitops join` to extract URLs into vars and generate the template

### Example

**wikis.yaml.template** (tracked, shared):
```yaml
wikis:
  - id: main
    url: "{{wiki_url_main}}"
    name: My Wiki
  - id: docs
    url: "{{wiki_url_docs}}"
    name: Documentation
```

**hosts/prod/vars.yaml** (encrypted, per-host):
```yaml
wiki_url_main: wiki.example.com
wiki_url_docs: docs.example.com
```

**hosts/dev-alice/vars.yaml** (encrypted, per-host):
```yaml
wiki_url_main: localhost
wiki_url_docs: localhost/docs
```

On `gitops pull`, `config/wikis.yaml` is rendered from the template + host vars, then the Caddyfile is regenerated as usual.

cicalese:
## Summary

- During `gitops init`, extract host-specific wiki URLs from `config/wikis.yaml` into `wikis.yaml.template` with `{{wiki_url_<id>}}` placeholders
- The actual `config/wikis.yaml` is gitignored and rendered from the template + per-host vars on `gitops join` and `gitops pull`
- Follows the same pattern already used for `.env` / `env.template`
- Existing repos get `config/wikis.yaml` added to `.gitignore` automatically on next push (backfill via `requiredGitignoreEntries`)
- Updated gitops documentation to cover wiki URL templating
- Added integration test verifying template round-trip (placeholders in template, URLs in vars, gitignore, tracking)

This prevents `gitops push` from overwriting production wiki URLs when developing locally against a restored backup.

Fixes #642

## Test plan

- [x] Unit tests pass (`go test ./internal/gitops/...`) — round-trip extraction + rendering, empty wikis edge case
- [x] Integration test verifies `wikis.yaml.template` is created during `gitops init`, contains placeholders, vars capture URLs, gitignore works
- [x] Manual: `gitops init` on a wiki farm creates `wikis.yaml.template` with placeholders and adds URL vars to `hosts/<name>/vars.yaml`
- [x] Manual: `gitops join` on a second host renders `config/wikis.yaml` with the local host's URLs from vars
- [x] Manual: `gitops pull` re-renders `config/wikis.yaml` from template, flags restart if content changed
- [x] Backward compatible: repos without `wikis.yaml.template` continue to work (graceful no-op)

### DIFF
--- a/cmd/gitops/init.go
+++ b/cmd/gitops/init.go
@@ -184,7 +184,25 @@ func runInitBootstrap(installPath, hostName, role, repoURL, keyFile string, forc
 		return err
 	}
 
-	// 6. Read admin passwords and add to vars.
+	// 6. Extract wikis.yaml template.
+	wikisContent, err := gitops.LoadWikisYAML(installPath)
+	if err != nil {
+		return err
+	}
+	if wikisContent != "" {
+		wikisTemplate, wikisVars, err := gitops.ExtractWikisTemplate(wikisContent)
+		if err != nil {
+			return err
+		}
+		if err := gitops.SaveWikisTemplate(installPath, wikisTemplate); err != nil {
+			return err
+		}
+		for k, v := range wikisVars {
+			vars[k] = v
+		}
+	}
+
+	// 7. Read admin passwords and add to vars.
 	passwords, err := gitops.ReadAdminPasswords(installPath)
 	if err != nil {
 		logging.Print(fmt.Sprintf("Warning: could not read admin passwords: %v\n", err))
@@ -193,7 +211,7 @@ func runInitBootstrap(installPath, hostName, role, repoURL, keyFile string, forc
 		vars["admin_password_"+wikiID] = password
 	}
 
-	// 7. Create hosts.yaml.
+	// 8. Create hosts.yaml.
 	// Use the Canasta ID from the installation registry if available.
 	canastaID := filepath.Base(installPath)
 	details, detailsErr := config.GetDetails(canastaID)
@@ -214,17 +232,17 @@ func runInitBootstrap(installPath, hostName, role, repoURL, keyFile string, forc
 		return err
 	}
 
-	// 8. Save vars.yaml.
+	// 9. Save vars.yaml.
 	if err := gitops.SaveVars(installPath, hostName, vars); err != nil {
 		return err
 	}
 
-	// 9. Save local host identity.
+	// 10. Save local host identity.
 	if err := gitops.SaveLocalHost(installPath, hostName); err != nil {
 		return err
 	}
 
-	// 10. Convert extensions and skins to submodules.
+	// 11. Convert extensions and skins to submodules.
 	if err := convertToSubmodules(installPath, "extensions"); err != nil {
 		fmt.Printf("Warning: could not convert extensions to submodules: %v\n", err)
 	}
@@ -232,7 +250,7 @@ func runInitBootstrap(installPath, hostName, role, repoURL, keyFile string, forc
 		fmt.Printf("Warning: could not convert skins to submodules: %v\n", err)
 	}
 
-	// 11. Initial commit.
+	// 12. Initial commit.
 	if err := gitops.AddAll(installPath); err != nil {
 		return err
 	}
@@ -241,7 +259,7 @@ func runInitBootstrap(installPath, hostName, role, repoURL, keyFile string, forc
 		return err
 	}
 
-	// 12. Add remote and push.
+	// 13. Add remote and push.
 	if err := gitops.AddRemote(installPath, "origin", repoURL); err != nil {
 		return err
 	}
@@ -495,6 +513,7 @@ var requiredGitignoreEntries = []struct {
 	comment string
 }{
 	{"config/backup/", "# Database dumps created by canasta backup"},
+	{"config/wikis.yaml", "# Rendered from wikis.yaml.template — host-specific URLs"},
 }
 
 // ensureGitignoreEntries appends any missing required entries to the repo's
--- a/cmd/gitops/join.go
+++ b/cmd/gitops/join.go
@@ -133,6 +133,21 @@ func runInitJoin(installPath, hostName, role, repoURL, keyFile string) error {
 			strings.Join(missing, "=... ")+"=...")
 	}
 
+	// Extract wiki URLs from the local wikis.yaml into vars.
+	wikisContent, err := gitops.LoadWikisYAML(installPath)
+	if err != nil {
+		return err
+	}
+	if wikisContent != "" {
+		_, wikisVars, err := gitops.ExtractWikisTemplate(wikisContent)
+		if err != nil {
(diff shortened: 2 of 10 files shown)


yaronkoren:
This is conceptually a little confusing - if GitOps is not used, then wikis.yaml is an actual settings file that can be edited, directly or indirectly; but if GitOps is used, then it becomes just an auto-generated file that should basically be ignored? If that's true, then what happens if the use of GitOps is removed - will the wikis.yaml.template setup still work?

Instead of a separate template file, what about just allowing variables directly in wikis.yaml - could that work?

cicalese:
Reply from Claude:

**On the conceptual confusion:** This is the same pattern already used for `.env` / `env.template`, which was implemented before this PR. Without gitops, `.env` and `wikis.yaml` are edited directly. With gitops, both become auto-generated from their templates and are gitignored. The `wikis.yaml` templating just follows the established convention.

**On removing gitops:** If gitops is removed, `wikis.yaml` still exists with the correct rendered values and continues to work normally. The `wikis.yaml.template` file just sits there unused. No breakage.

**On using variables directly in wikis.yaml:** The core problem is git's working tree. If `wikis.yaml` contained placeholders in git but real values locally, then after every `gitops pull` or `gitops join` renders the file, `git status` would show `wikis.yaml` as modified — permanently. This makes it impossible to distinguish real changes from the expected template-vs-rendered difference, and `gitops status` / `gitops diff` become unreliable.

Possible workarounds (swap-and-restore on push, git clean/smudge filters, `--assume-unchanged`) all add significant complexity and fragility compared to the current approach. The separate template file keeps things clean: the template (with placeholders) is tracked in git, the rendered file (with real values) is gitignored, and `git status` is always accurate.

The "two files" pattern (tracked template + gitignored real file) is well-established — `.env` / `.env.example` is standard in most frameworks. The conceptual shift ("this file is now auto-generated under gitops") is real, but it already existed for `.env` before this PR.

yaronkoren:
Right, I wish I had noticed that .env.template thing before. :)

Just to be clear - for variables in wikis.yaml, I meant that the variables would remain there, not that they would get substituted in (which would indeed be a bad idea). Or is having that use of variables inefficient, since it would mean that multiple files would have to be read on every page load?

cicalese:
Reply from Claude:

The idea of keeping variables like `{{wiki_url_mywiki}}` in `wikis.yaml` permanently and resolving them at runtime is interesting, but the main issue is architectural: `wikis.yaml` is read by `FarmConfigLoader.php` at runtime, and PHP currently has zero knowledge of the gitops variable system. The variable values live in `hosts/<hostname>/vars.yaml` inside the gitops repo — a location PHP doesn't know about and shouldn't need to.

Making this work would mean:

1. Teaching PHP about gitops variables (coupling the MediaWiki runtime to CLI/gitops concerns)
2. Making the variable values available somewhere PHP can find them (a new file, or injecting them into the container environment)
3. Having two different variable resolution patterns — `.env` can't do this (docker-compose doesn't support variable indirection from external files), so `.env` would still need the template approach

It's a cleaner separation to keep variable resolution entirely in the CLI layer (which already owns the gitops workflow) rather than splitting it across CLI and PHP.

yaronkoren:
Alright. I suppose it's too late, since the architecture is already in place, but isn't it weird that the GitOps code has its own templating/variable system? Of course, this system is _helpful_ when using GitOps, but couldn't it be helpful in other circumstances as well? Couldn't two separate Canasta instances want to have variations on the same wikis.yaml file, even if you don't use GitOps? (I admit that using Git would be the ideal solution.)

I suppose I have a broader confusion, centered around Claude's statement that "The variable values live in `hosts/<hostname>/vars.yaml` inside the gitops repo". What is this gitops repo? And does it contain a bunch of YAML files? I thought the CLI GitOps commands were basically just wrappers around Git. (Should have looked at the patches more closely.)

cicalese:
(This is actually Cindy.)

> Alright. I suppose it's too late, since the architecture is already in place, but isn't it weird that the GitOps code has its own templating/variable system? Of course, this system is _helpful_ when using GitOps, but couldn't it be helpful in other circumstances as well? Couldn't two separate Canasta instances want to have variations on the same wikis.yaml file, even if you don't use GitOps? (I admit that using Git would be the ideal solution.)

If they are two separate Canasta instances, though, they will have separate config directories. I'm not at all clear on what the benefit of a template would be in that case without git.

> I suppose I have a broader confusion, centered around Claude's statement that "The variable values live in `hosts/<hostname>/vars.yaml` inside the gitops repo". What is this gitops repo?

The GitHub repo that the admin created for use with the gitops feature.

> And does it contain a bunch of YAML files? I thought the CLI GitOps commands were basically just wrappers around Git. (Should have looked at the patches more closely.)

Did you read the gitops guide, especially https://docs.canasta.wiki/guide/gitops/#repository-structure and the two sections right before that? Hopefully that will clarify what the repo contains and a bit more of the rationale. I also created https://notes.amethyst-ck.com/wiki/Setting_up_a_local_Canasta_dev_copy to communicate a development workflow to other team members on a wiki where there are multiple admins and this approach is being used to support the multi-editor/multi-wiki farm process.

Also, https://github.com/cicalese/test-gitops is a repo I set up to test the functionality (prior to this PR). It is public, but normally one would create such a repo as private.

yaronkoren:
Thanks for this explanation. These two guides (and your explanation here) are certainly helpful. Still, it looks to me like the Canasta documentation could make it clearer which files/directories are _specific_ to GitOps - I don't see that information on that page.

Anyway, given that this current setup works, and that you (and Claude) feel strongly that it's the right solution, this seems like the right approach to take.

yaronkoren (approved):

PR #649, opened 2026-03-14
Move --wiki flag to maintenance subcommands and iterate over all wikis in script

### LINKED ISSUE #648: maintenance script doesn't iterate over wikis; --wiki flag exposed on exec
When running `canasta maintenance script` on a wiki farm without specifying `--wiki`, the script runs without a wiki context and MediaWiki produces confusing error messages that don't indicate the actual problem.

Additionally, the `--wiki/-w` flag is defined as a persistent flag on the `maintenance` command, making it available on `exec` where it has no meaning.

### Changes

1. Move `--wiki/-w` from a persistent flag on `maintenance` to a local flag on each subcommand that uses it (`update`, `script`, `extension`). The `exec` subcommand no longer shows `--wiki`.
2. Make `script` iterate over all wikis when `--wiki` is omitted, matching the existing behavior of `update` and `extension`.

cicalese:
Closes #648

## Summary

- Move `--wiki/-w` from a persistent flag on `maintenance` to a local flag on `update`, `script`, and `extension` — `exec` no longer shows `--wiki`
- Make `script` iterate over all wikis when `--wiki` is omitted, matching the behavior of `update` and `extension`
- Updated tests to verify wiki iteration behavior

## Test plan

- [x] Unit tests pass (`go test ./cmd/maintenance/...`)
- [x] `canasta maintenance exec --help` does not show `--wiki`

### DIFF
--- a/cmd/maintenance/exec.go
+++ b/cmd/maintenance/exec.go
@@ -15,8 +15,9 @@ func newExecCmd(instance *config.Installation) *cobra.Command {
 	var service string
 
 	cmd := &cobra.Command{
-		Use:   "exec [command ...]",
-		Short: "Execute a command in a running container",
+		Use:                   "exec [flags] [command ...]",
+		DisableFlagsInUseLine: true,
+		Short:                 "Execute a command in a running container",
 		Long: `Execute a command or open an interactive shell in a running container
 of a Canasta installation.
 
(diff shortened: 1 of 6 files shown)


yaronkoren (approved):
This feels like three patches in one, but they're all good!

cicalese:
I accidentally committed the next PR's changes to this PR then reverted. You may have looked between the commit and the revert. I see that the PR message was not fixed, but I'll fix it now.

PR #653, opened 2026-03-15
Standardize flags-first argument style in maintenance subcommands

### LINKED ISSUE #652: Standardize flags-first argument style in maintenance subcommands
## Problem

The maintenance subcommands (`script`, `exec`, `extension`) handle flags and positional arguments inconsistently:

- **`extension`** correctly uses `SetInterspersed(false)` so flags must come before positional arguments, but Cobra's auto-generated usage line shows `[flags]` at the end, which is misleading.
- **`script`** requires the script name and arguments to be quoted as a single argument. It could use `SetInterspersed(false)` like `extension` to allow unquoted arguments.
- **`exec`** requires `--` to separate flags from the container command. With `SetInterspersed(false)`, the `--` would be unnecessary in most cases.

With `SetInterspersed(false)`, flags placed after the first positional argument are silently treated as positional arguments, so `[flags]` at the end of the usage line is demonstrably wrong.

## Proposed fix

1. Add `SetInterspersed(false)` to `script` and `exec`, matching `extension`
2. Use `DisableFlagsInUseLine: true` and put `[flags]` before positional args in the `Use` string for all three
3. Update `script` to accept `ArbitraryArgs` instead of a single quoted argument
4. Update `exec` examples to show the simpler form without `--`
5. Update long descriptions and examples to document flags-first style consistently

cicalese:
Closes #652

## Summary

- Add `SetInterspersed(false)` to `canasta maintenance script` and `canasta maintenance exec`, matching `canasta maintenance extension`'s existing behavior — flags must come before positional arguments
- Use `DisableFlagsInUseLine` with `[flags]` placed before positional args in the `Use` string for all three `canasta maintenance` subcommands, accurately reflecting the required ordering
- `canasta maintenance script` no longer requires quotes around the script name and arguments (uses `ArbitraryArgs` instead of `RangeArgs(0, 1)`)
- `canasta maintenance exec` no longer requires `--` to separate flags from the container command
- Updated long descriptions and examples for all three subcommands

## Test plan

- [x] Unit tests pass (`go test ./cmd/maintenance/...`)
- [x] Manual: `canasta maintenance script --help` shows `[flags]` before positional args
- [x] Manual: `canasta maintenance exec --help` shows `[flags]` before positional args
- [x] Manual: `canasta maintenance extension --help` shows `[flags]` before positional args

### DIFF
--- a/cmd/maintenance/exec.go
+++ b/cmd/maintenance/exec.go
@@ -15,14 +15,18 @@ func newExecCmd(instance *config.Installation) *cobra.Command {
 	var service string
 
 	cmd := &cobra.Command{
-		Use:   "exec [-- command ...]",
-		Short: "Execute a command in a running container",
+		Use:                   "exec [flags] [command ...]",
+		DisableFlagsInUseLine: true,
+		Short:                 "Execute a command in a running container",
 		Long: `Execute a command or open an interactive shell in a running container
 of a Canasta installation.
 
 With no arguments and no --service flag, lists the running services.
 With --service (or -s) and no command, opens an interactive bash shell.
-With --service and a command after --, runs that command.
+With --service and a command, runs that command.
+
+Flags (-i, -s) must come before the command. Everything after the first
+non-flag argument is treated as the command and its arguments.
 
 The default service is "web" (the MediaWiki container).`,
 		Example: `  # List running services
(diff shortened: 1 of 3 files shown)


yaronkoren (approved):
Great syntax improvement!

PR #655, opened 2026-03-16
Fix Varnish caching: remove 48h TTL override and add no-store for browsers

### LINKED ISSUE #654: Fix Varnish caching: remove 48h TTL override and add no-store for browsers
## Problem

Two Varnish caching issues cause anonymous users to see stale content:

### 1. 48-hour TTL override ignores MediaWiki's intended cache duration

The current `vcl_backend_response` contains:

```vcl
if (beresp.ttl < 48h) {
    set beresp.ttl = 48h;
}
```

MediaWiki sets `s-maxage=18000` (5 hours) via `$wgCdnMaxAge`, but this VCL block silently overrides Varnish's internal TTL to 48 hours for all responses. This means if a purge is missed, Varnish serves stale content for up to 48 hours instead of 5.

This appears to have originated as an overcorrection for a problematic line in the [MediaWiki reference VCL](https://www.mediawiki.org/wiki/Manual:Varnish_caching) that marked anything with TTL under 48h as uncacheable. Instead of removing the block, it was changed to force a 48h TTL — the opposite extreme.

### 2. Browser bfcache serves stale content on back/forward navigation

When a user visits a page, the browser stores a bfcache (back-forward cache) snapshot. This snapshot is **completely separate from HTTP cache** and ignores `Cache-Control` headers including `max-age=0` and `must-revalidate`. When the user navigates back, the browser restores the snapshot without making any request to Varnish.

This means:
- After a purge, hitting back shows the old content
- Other users who previously visited the page and navigate back to it see stale content regardless of what Varnish has

## Fix

1. **Remove the 48h TTL override** — let Varnish respect the TTL MediaWiki sends via `s-maxage`
2. **Add `vcl_deliver`** — set `Cache-Control: no-store` on responses that contain `s-maxage`, which disqualifies the page from bfcache while leaving API responses, images, and logged-in user pages unaffected. By the time `vcl_deliver` runs, Varnish has already recorded its internal TTL from `s-maxage`, so overwriting the header only affects what gets sent to the browser.

cicalese:
## Summary

- Remove the `vcl_backend_response` block that forced all responses to cache for 48 hours, ignoring MediaWiki's `s-maxage=18000` (5 hours) setting
- Add `vcl_deliver` to set `Cache-Control: no-store` on publicly cacheable responses, preventing browser bfcache from serving stale content on back/forward navigation

## Details

The 48h TTL override appears to have originated as an overcorrection for a problematic line in the [MediaWiki reference VCL](https://www.mediawiki.org/wiki/Manual:Varnish_caching) that marked anything with TTL under 48h as uncacheable. Instead of removing the block, it was changed to force a 48h TTL — the opposite extreme.

The `vcl_deliver` block scopes `no-store` to responses containing `s-maxage` (i.e., anonymous page views where `$wgUseCdn = true`), leaving API responses, images, and logged-in user pages unaffected. By the time `vcl_deliver` runs, Varnish has already recorded its internal TTL from `s-maxage`, so overwriting the header only affects what the browser sees.

Fixes #654

## Test plan

- [x] Verify Varnish still caches anonymous page views (`curl -sI` should show `Cache-Control: no-store` instead of `s-maxage=18000, must-revalidate, max-age=0`)
- [x] Edit a page and confirm anonymous users see the update without purging (within the 5-hour TTL window)
- [x] Visit a page, purge it, hit back — confirm the back button shows updated content (bfcache fix)
- [x] Confirm logged-in user pages, API responses, and images are unaffected


yaronkoren (approved):
Great! I hope this fixes the caching issue!

PR #658, opened 2026-03-16
Add tests for EnsureRunning function in mockOrchestrator

th33k:
This pull request adds and improves tests for the `EnsureRunning` function in the orchestrator package, focusing on verifying its behavior under different container running states. It also updates the `mockOrchestrator` to better support these test scenarios.

**Testing improvements:**

* Added three new tests to `orchestrator_test.go` to verify `EnsureRunning` behavior when containers are already running, not running, and when starting containers fails.

**Test infrastructure updates:**

* Updated the `mockOrchestrator` struct to include a `checkRunningErr` field, allowing tests to simulate different running states. The `CheckRunningStatus` method now returns this error. [[1]](diffhunk://#diff-6d356b25a011945bcba430d963ba8f9428d637885ec25c246dc990d59d750217R19) [[2]](diffhunk://#diff-6d356b25a011945bcba430d963ba8f9428d637885ec25c246dc990d59d750217L55-R56)

### DIFF
--- a/internal/orchestrators/orchestrator_test.go
+++ b/internal/orchestrators/orchestrator_test.go
@@ -53,6 +53,7 @@ func (m *mockOrchestrator) ExecStreaming(_, _, _ string) error {
 }
 
 func (m *mockOrchestrator) CheckRunningStatus(_ config.Installation) error {
+	m.calls = append(m.calls, "CheckRunningStatus")
 	return m.checkRunningErr
 }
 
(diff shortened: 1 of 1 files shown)


cicalese (commented):
Looks good — the tests cover all three code paths through `EnsureRunning` and the mock change is clean. One minor suggestion:

`CheckRunningStatus` doesn't record the call in `m.calls`. For consistency with `Start`, `Stop`, etc., consider adding `m.calls = append(m.calls, "CheckRunningStatus")`. Then `TestEnsureRunningAlreadyRunning` could also assert that `CheckRunningStatus` *was* called, not just that `Start` wasn't.

th33k:
> Looks good — the tests cover all three code paths through `EnsureRunning` and the mock change is clean. One minor suggestion:
> 
> `CheckRunningStatus` doesn't record the call in `m.calls`. For consistency with `Start`, `Stop`, etc., consider adding `m.calls = append(m.calls, "CheckRunningStatus")`. Then `TestEnsureRunningAlreadyRunning` could also assert that `CheckRunningStatus` _was_ called, not just that `Start` wasn't.

@cicalese,

Thanks for the suggestion. I’ve updated `mockOrchestrator` to record `CheckRunningStatus` calls in `m.calls` for consistency. Also added an assertion in `TestEnsureRunningAlreadyRunning` to verify it’s being called.

cicalese (approved):

PR #685, opened 2026-03-21
fix: skip wikis where extension is not loaded in maintenance extension

### LINKED ISSUE #684: maintenance extension fails instead of skipping wikis where extension is not loaded
## Description

When running `canasta maintenance extension <ExtName> script.php` without `-w` in a wiki farm, the command iterates over all wikis. For each wiki, it checks whether the extension is loaded. If the extension is not loaded on a particular wiki, it returns an error and stops — it does not skip that wiki and continue to the next.

This means if an extension is selectively enabled (e.g., only on the `docs` wiki but not on `main`), running the maintenance script without `-w` will fail on the first wiki where it's not loaded, even though the intent is to run it everywhere it applies.

## Expected behavior

The command should skip wikis where the extension is not loaded (with an informational message) and continue to the remaining wikis, rather than failing with an error.

## Steps to reproduce

1. Create a wiki farm with two wikis: `main` and `docs`
2. Enable an extension (e.g., SemanticMediaWiki) only on `docs`
3. Run `canasta maintenance extension SemanticMediaWiki rebuildData.php` without `-w`
4. The command fails with: `extension "SemanticMediaWiki" is not loaded for wiki "main"`

## Relevant code

`cmd/maintenance/extension.go`, `runExtensionScript` function (line 249):
```go
if !loaded {
    return fmt.Errorf("extension %q is not loaded for wiki %q", extName, checkWiki)
}
```

cicalese:
## Summary

- When running `canasta maintenance extension <ext> <script>` without `-w` in a wiki farm, the command now skips wikis where the extension is not loaded (with an informational message) instead of failing with an error
- When `-w` is explicitly provided, the error is still returned since the user specifically targeted that wiki
- If the extension is not loaded on any wiki, a clear error is returned

Fixes #684

## Test plan

- [x] `TestRunExtensionScriptMultiWikiSkip/skips_wiki_where_extension_not_loaded` — skips wiki where extension is not loaded, runs on the other
- [x] `TestRunExtensionScriptMultiWikiSkip/runs_on_all_wikis_where_extension_is_loaded` — runs on all wikis when loaded everywhere
- [x] `TestRunExtensionScriptMultiWikiSkip/not_loaded_on_any_wiki` — no script runs when extension not loaded anywhere
- [x] `TestRunExtensionScriptMultiWikiSkip/single_wiki_where_extension_is_loaded` — works for single-wiki installations
- [x] `TestRunExtensionScriptNotLoadedSentinel/extension_not_in_loaded_list` — returns `errExtNotLoaded` sentinel
- [x] `TestRunExtensionScriptNotLoadedSentinel/empty_loaded_list` — returns `errExtNotLoaded` for empty extension list
- [x] Manual: enabled Cargo only on `docs` wiki in a 2-wiki farm, ran without `-w` → skipped `main`, ran on `docs`
- [x] Manual: ran with `-w main` → errored as expected ("not loaded for wiki main")
- [x] Manual: ran with `-w docs` → succeeded

### DIFF
--- a/cmd/maintenance/extension.go
+++ b/cmd/maintenance/extension.go
@@ -1,6 +1,7 @@
 package maintenance
 
 import (
+	"errors"
 	"fmt"
 	"path/filepath"
 	"regexp"
@@ -14,6 +15,10 @@ import (
 	"github.com/CanastaWiki/Canasta-CLI/internal/farmsettings"
 	"github.com/CanastaWiki/Canasta-CLI/internal/orchestrators"
 )
+
+// errExtNotLoaded is returned by runExtensionScript when the extension
+// is not loaded for a particular wiki.
+var errExtNotLoaded = errors.New("extension not loaded")
 
 // wikiArgRe matches --wiki=value or --wiki value in a script argument string.
 var wikiArgRe = regexp.MustCompile(`(?:^|\s)--wiki[=\s](\S+)`)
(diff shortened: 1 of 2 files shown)


yaronkoren (approved):
This is a great improvement!

Issue #616, opened 2026-03-10
Test: add unit tests for consolidateUserDir


cicalese:
## Summary

Add unit tests for the `consolidateUserDir` function in `internal/devmode/devmode.go`.

## What to test

This function consolidates user-installed extensions or skins from their real directory into the extracted code directory, then replaces the original with a symlink.

Test cases:
- User directory exists with contents — should copy contents and create symlink
- User directory does not exist — should be a no-op
- Symlink already exists — should handle gracefully
- Target directory already exists in code dir

## How to test

Use `t.TempDir()` to create temporary directory structures with files and symlinks. No Docker or orchestrator needed.

## Context

Part of the test coverage initiative (#241).

cicalese:
Done in #711.

PR #711, opened 2026-03-23
test: add unit tests for consolidateUserDir function

th33k:
This pull request adds comprehensive tests for the `consolidateUserDir` function in the `internal/devmode/devmode_test.go` file. The new tests cover various scenarios related to user directory consolidation and symlink behavior, with special handling to skip symlink tests on Windows. Additionally, a helper function `copyDir` is introduced to support directory copying in the test environment.

**Testing improvements:**

* Added `TestConsolidateUserDir` to verify the behavior of `consolidateUserDir` in multiple scenarios, including when directories exist, do not exist, symlinks are present, and target directories already exist. The tests ensure correct file copying and symlink creation, and skip symlink tests on Windows for reliability.

**Test utilities:**

* Introduced a `copyDir` helper function to recursively copy directory contents during tests, supporting the simulation of the `cp -r` command.
* Mocked the `execute.Run` function within tests to intercept and simulate external command execution, restoring the original function after the tests complete.

**Dependency updates:**

* Imported additional packages (`io`, `runtime`, and `github.com/CanastaWiki/Canasta-CLI/internal/execute`) to support new test logic and utilities.

Fix issue https://github.com/CanastaWiki/Canasta-CLI/issues/616

### DIFF
--- a/internal/devmode/devmode_test.go
+++ b/internal/devmode/devmode_test.go
@@ -1,9 +1,13 @@
 package devmode
 
 import (
+	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
+
+	"github.com/CanastaWiki/Canasta-CLI/internal/execute"
 )
 
 func TestIsDevModeSetup(t *testing.T) {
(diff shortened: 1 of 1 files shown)


yaronkoren:
@th33k - it's great that you're doing these microtasks. But please don't do any more after this one! We're trying to limit applicants to 2-3 microtasks each. (Really it should have been 1 each, but it's too late for that.)

th33k:
> @th33k - it's great that you're doing these microtasks. But please don't do any more after this one! We're trying to limit applicants to 2-3 microtasks each. (Really it should have been 1 each, but it's too late for that.)

Got it, thanks @yaronkoren! I’ll stop here and focus on my proposal.

cicalese (approved):
Thanks for the contribution! Good coverage of the consolidation paths, especially the overwrite case. One minor note: the symlink target in the "already exists" test uses `filepath.Join("..", targetDir)` with an absolute path, which creates an unusual path — but it correctly tests the idempotent behavior.

PR #713, opened 2026-03-23
test: add unit tests for migrateDirectoryStructure

### LINKED ISSUE #612: Test: add unit tests for migrateDirectoryStructure migration
## Summary

Add unit tests for the `migrateDirectoryStructure` function in `cmd/upgrade/upgrade.go`.

## What to test

This function migrates legacy directory layouts: renames `config/<wiki_id>/` to `config/settings/wikis/<wiki_id>/` and moves global PHP files to `config/settings/global/`.

Test cases:
- Legacy layout present — should migrate directories
- Already-migrated layout — should be a no-op
- Multiple wiki directories
- Dry-run mode — should report changes without modifying anything
- Mixed state (some wikis migrated, some not)

## How to test

Use `t.TempDir()` to create temporary directory structures. No Docker or orchestrator needed.

## Context

Part of the test coverage initiative (#241).

kajal-jotwani:
The pull request adds unit tests for the `migrateDirectoryStructure` function in `cmd/upgrade/upgrade.go`.

The function has two migration paths (per-wiki dirs and global PHP files) so I tried to make sure both are covered properly, including the edge cases:

- basic legacy layout where both a wiki dir and a global PHP file need moving
- already-migrated instance - this one also puts a legacy PHP file in `config/settings/` alongside its copy in `global/` so the skip-if-destination-exists branch actually gets hit, not just skipped over silently
- multiple wikis in legacy layout
- dry-run - reports changes but touches nothing on disk
- mixed state where one wiki is legacy, one is already migrated, and one has no directory at all

Fixes: #612

### DIFF
--- a/cmd/upgrade/upgrade_test.go
+++ b/cmd/upgrade/upgrade_test.go
@@ -354,3 +354,269 @@ func TestRemoveEmptyComposerLocalPreservesPopulated(t *testing.T) {
 		t.Error("file content should be unchanged")
 	}
 }
+
+func writeWikisYAML(t *testing.T, installPath string, wikiIDs ...string) {
+	t.Helper()
+
+	configDir := filepath.Join(installPath, "config")
+	if err := os.MkdirAll(configDir, 0755); err != nil {
+		t.Fatal(err)
+	}
+
+	var b strings.Builder
+	b.WriteString("wikis:\n")
+	for _, id := range wikiIDs {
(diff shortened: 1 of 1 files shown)


cicalese (approved):
Thanks for the contribution! Thorough coverage — the mixed-state and dry-run tests are especially valuable for catching real-world upgrade edge cases.

Issue #716, opened 2026-03-25
Integration test coverage not collected in CI


cicalese:
## Problem

Integration test coverage data is not being collected in CI due to two issues:

1. **Build target uses file path instead of package path**: `go build -cover` only applies coverage instrumentation when building a package (e.g., `github.com/CanastaWiki/Canasta-CLI`), not when building a single file (`../../canasta.go`). The current build target `../../canasta.go` produces a binary with no coverage instrumentation.

2. **Coverage output path is relative**: `INTEGRATION_COVER_PROFILE=integration-coverage.out` resolves relative to the test binary's working directory (`tests/integration/`), not the repo root. Subsequent workflow steps that look for the file at the repo root can't find it.

Additionally, `config.Add()` panics with "assignment to entry in nil map" when `conf.json` exists but is empty or has a null `Instances` field, which was discovered while testing these fixes.

## Expected behavior

- Integration coverage data should be collected and uploaded as a CI artifact
- Combined (unit + integration) coverage should be reported in the step summary
- `config.Add()` should handle nil maps gracefully

PR #720, opened 2026-03-26
improve: clarify coverage summary and artifact names

cicalese:
## Summary
- Use a markdown table for the coverage summary so percentages are clearly labeled by scope
- Rename artifacts from `unit-coverage` / `integration-coverage` / `combined-coverage` to `coverage-profile-*` to avoid confusion with the percentage values (the file sizes in KB were visually similar to coverage percentages)

## Test plan
- [ ] Coverage summary renders as a table with Unit and Combined rows
- [ ] Artifacts appear as `coverage-profile-unit`, `coverage-profile-integration`, `coverage-profile-combined`

Relates to #716


yaronkoren (approved):
Cool, this will be a fancy printout!

PR #722, opened 2026-03-26
Remove unused Orchestrators field from conf.json

Manvi-9791:
Removes the unused Orchestrators entry from conf.json handling in internal/config.
Deletes obsolete orchestrator config APIs: AddOrchestrator, GetOrchestrator, and OrchestratorExists.
Stops initializing an empty Orchestrators map in ensureInitialized.
Updates Compose runtime behavior to resolve the compose binary at runtime instead of reading it from config.
Removes tests that only covered the deleted orchestrator config APIs.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -25,12 +25,7 @@ type Instance struct {
 	BuildFrom      string `json:"buildFrom,omitempty"`
 }
 
-type Orchestrator struct {
-	ID, Path string
-}
-
 type Canasta struct {
-	Orchestrators       map[string]Orchestrator
 	Instances           map[string]Instance `json:"Instances"`
 	LegacyInstallations map[string]Instance `json:"Installations,omitempty"`
 }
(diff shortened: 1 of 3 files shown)


cicalese (commented):
Thanks for this PR! The config cleanup looks great.

The linter flagged an issue in the `compose.go` changes, and looking at it more closely, `getCompose()` is effectively dead code — `runCompose` and `composeCommand` already fall back to `docker compose` when `Path` is empty, and the standalone `docker-compose` binary doesn't exist on most modern Docker installs, so the `LookPath` call almost always returns an empty path.

The right fix is to remove `getCompose()` entirely, drop the `compose` parameter from `runCompose` and `composeCommand`, and have them always use `docker compose`. That touches ~15 call sites though, so let's handle it separately. Please revert the `compose.go` changes from this PR and we'll do the compose simplification as a follow-up.

cicalese:
Thanks for the cleanup effort. After looking more closely, the `Orchestrator` type and `GetOrchestrator`/`AddOrchestrator` functions are still actively used by `compose.go` — they're not dead code, even though the `Orchestrators` map in `conf.json` is never populated in practice. Removing them requires refactoring `compose.go`, which is a larger change than this PR's scope. Closing for now.

cicalese:
Apologies for closing this prematurely — after further review, the compose.go refactor is necessary to cleanly remove the Orchestrator type, and the Printf → Fprintf fix is a legitimate bug fix. Reopening and merging. We'll follow up with a separate PR to simplify the compose.go changes (dropping the LookPath in favor of always using `docker compose` directly).

cicalese (approved):

PR #723, opened 2026-03-31
test: add tests for IsElasticsearchEnabled and IsVarnishEnabled

devjadiya:
## What this PR does

Adds 20 unit tests covering `IsElasticsearchEnabled` and `IsVarnishEnabled` in `internal/canasta/canasta.go` — both functions currently have zero test coverage despite being called in production paths.

Addresses: T418757 (related)

## Key finding that motivated this PR

The three feature flag functions have inconsistent defaults when the env key is absent:

| Function | Key absent → returns |
|---|---|
| `IsObservabilityEnabled` | `false` |
| `IsElasticsearchEnabled` | `false` |
| `IsVarnishEnabled` | **`true`** ← backward compat |

`IsVarnishEnabled` defaults to `true` when the key is absent or empty, for backward compatibility with deployments that ran Varnish before the flag existed. The other two default to `false`. This asymmetry is a silent regression trap — any future refactoring that normalizes the three functions would break Varnish behavior for existing deployments. These tests document and enforce that contract.

## Tests added (20 total)

**`TestIsElasticsearchEnabled` — 9 cases:**
- `true` / `TRUE` / `True` (case-insensitive)
- `false` / `FALSE`
- key absent → false (default disabled)
- key set to empty string → false
- non-boolean value `"yes"` → false
- `"1"` is not treated as true → false

**`TestIsVarnishEnabled` — 11 cases:**
- `true` / `TRUE` / `True` (case-insensitive)
- `false` / `FALSE`
- key absent → **true** (backward compatibility default)
- key set to empty string → **true** (same default-true path)
- non-boolean value `"yes"` → false
- `"1"` is not treated as true → false
- unrelated env keys do not affect the result

## How to run
`go test ./internal/canasta/... -run "TestIsElasticsearch|TestIsVarnish" -v`

### DIFF
--- a/internal/canasta/canasta_test.go
+++ b/internal/canasta/canasta_test.go
@@ -1147,3 +1147,147 @@ func TestGetBaseImage(t *testing.T) {
 		}
 	})
 }
+
+// Tests for IsElasticsearchEnabled and IsVarnishEnabled.
+
+func TestIsElasticsearchEnabled(t *testing.T) {
+	tests := []struct {
+		name    string
+		envVars map[string]string
+		want    bool
+	}{
+		{
+			name:    "explicitly true lowercase",
+			envVars: map[string]string{"CANASTA_ENABLE_ELASTICSEARCH": "true"},
+			want:    true,
+		},
+		{
+			name:    "explicitly true uppercase",
+			envVars: map[string]string{"CANASTA_ENABLE_ELASTICSEARCH": "TRUE"},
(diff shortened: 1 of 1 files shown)


cicalese (approved):
Thanks for the contribution! The tests are well-structured and the comments documenting the asymmetric default behavior between Varnish and Elasticsearch are a nice touch — they serve as living documentation.

PR #712, opened 2026-03-23
test: add unit tests for purgeFromLine

kajal-jotwani:
The pull request adds unit tests for the `purgeFromLine` function in `cmd/backup/schedule.go`.

As mentioned in the issue the tests covers:
- Unquoted value (`30d`)
- Double-quoted value (`"30d"`)
- Single-quoted value (`'30d'`)
- No flag present
- Multiple flags including `--older-than`
- Empty line

```
kajal@kajal-HP-Pavilion-Laptop-15-eh2xxx:~/Canasta-CLI$ go test ./cmd/backup/... -run TestPurgeFromLine -v
=== RUN   TestPurgeFromLine
=== RUN   TestPurgeFromLine/unquoted_value
=== RUN   TestPurgeFromLine/double-quoted_value
=== RUN   TestPurgeFromLine/single-quoted_value
=== RUN   TestPurgeFromLine/no_flag_present
=== RUN   TestPurgeFromLine/multiple_flags_including_older-than
=== RUN   TestPurgeFromLine/empty_line
--- PASS: TestPurgeFromLine (0.00s)
    --- PASS: TestPurgeFromLine/unquoted_value (0.00s)
    --- PASS: TestPurgeFromLine/double-quoted_value (0.00s)
    --- PASS: TestPurgeFromLine/single-quoted_value (0.00s)
    --- PASS: TestPurgeFromLine/no_flag_present (0.00s)
    --- PASS: TestPurgeFromLine/multiple_flags_including_older-than (0.00s)
    --- PASS: TestPurgeFromLine/empty_line (0.00s)
PASS
ok      github.com/CanastaWiki/Canasta-CLI/cmd/backup   
```

Fixes: #614

PR #728, opened 2026-04-08
Simplify compose abstraction: remove LookPath, always use docker compose

cicalese:
## Summary
- Remove `composeBinary` struct and `getCompose()` LookPath discovery from #722
- Simplify `runCompose`, `composeCommand`, `getComposeImages` to always use `docker compose` directly
- Canasta requires `docker compose` (the plugin), not standalone `docker-compose`

Fixes #727

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./...` passes

PR #730, opened 2026-04-08
Fix gofmt: add missing trailing newlines in test files

cicalese:
## Summary
Add missing trailing newlines in test files merged from #711, #712, #713.

## Test plan
- [x] `gofmt -l` reports no issues


internal/orchestrators/compose.go
```diff
@@ -33,6 +33,8 @@ type composeBinary struct {
 }
 
 // getCompose discovers the compose binary at runtime.
+//
+//nolint:unparam // error return removed in #728
```
yaronkoren, line 37: Is this meant to be here? It seems unrelated to the rest.

cicalese: Yes. A contributed PR added this line. It will be removed in #728 shortly, but the nolint is necessary to make lint pass for now.

yaronkoren (approved):

PR #736, opened 2026-04-13
Fix upgrade overriding user-set CANASTA_IMAGE

### LINKED ISSUE #735: `canasta upgrade` overrides manual canasta image config set 3.5.4 back to 3.5.2
CANASTA_IMAGE was set to 3.5.2; updated with 
`canasta config set CANASTA_IMAGE=ghcr.io/canastawiki/canasta:3.5.4` 
which updated my wiki to 1.43.8 as it should. Then ran canasta upgrade, and it reverted the setting. Output: 
>CLI is up to date (v3.6.2)
>...
>Checking for config migrations...
>  Set CANASTA_IMAGE=ghcr.io/canastawiki/canasta:3.5.2 in .env
>Migrations applied.

That also reverts wiki to 1.43.6 (6f1ddad)

cicalese:
## Summary

`backfillCanastaImage` was overwriting `CANASTA_IMAGE` with the CLI's built-in default on every upgrade, even when the user had intentionally set a different version via `canasta config set`. Now it only sets the value when the key is **missing** from `.env` (the original backfill intent for pre-CANASTA_IMAGE instances).

Fixes #735

## Test plan

- [x] Unit tests updated and passing
- [ ] `canasta config set CANASTA_IMAGE=ghcr.io/canastawiki/canasta:3.5.4` followed by `canasta upgrade` preserves 3.5.4
- [ ] Fresh instance without CANASTA_IMAGE in .env gets it backfilled on upgrade

### DIFF
--- a/cmd/upgrade/upgrade.go
+++ b/cmd/upgrade/upgrade.go
@@ -648,10 +648,11 @@ func removeLegacyGitDir(installPath string, dryRun bool) (bool, error) {
 	return true, nil
 }
 
-// backfillCanastaImage ensures CANASTA_IMAGE in .env matches the current
-// CLI's default image tag. This handles both instances that predate the
-// CANASTA_IMAGE variable and instances where the CLI was upgraded to a
-// newer version but the image tag was not updated.
+// backfillCanastaImage ensures CANASTA_IMAGE is present in .env for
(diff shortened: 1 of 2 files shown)


yaronkoren (approved):
Looks good!

PR #740, opened 2026-04-25
Add Go-to-Ansible bridge program for v3.7.0 release

### LINKED ISSUE #739: Add Go-to-Ansible bridge program for v3.7.0 release
Part of the Canasta-Ansible v4.0.0 release plan (see [`canasta-ansible-release-plan.md`](https://github.com/CanastaWiki/Canasta-Ansible/blob/main/canasta-ansible-release-plan.md)). v3.7.0 will be the final Canasta-CLI Go release. After that, this repo is renamed to `Canasta-Go` (archived) and the canonical `canasta` CLI becomes the Ansible-based implementation under the `Canasta-CLI` repo name.

To preserve the upgrade path for users running existing Canasta-CLI Go binaries (whose `canasta upgrade` selfupdates from `github.com/CanastaWiki/Canasta-CLI/releases/latest`), this PR adds a minimal standalone Go program that prints the migration notice and exits regardless of arguments. Cross-compiled binaries from this program will be attached as release assets to v3.7.0 (this repo) and to every Canasta-CLI 4.x.x release.

cicalese:
## Summary
Adds a minimal standalone Go program that prints a migration notice directing users to install the Ansible-based Canasta CLI 4.0.0+ via `get.canasta.wiki`. Any invocation of the bridge prints the message — there's no half-functional state.

No existing Cobra code is modified; `cmd/upgrade/upgrade.go`, `internal/selfupdate`, `build.sh`, and the rest of the CLI implementation remain unchanged in this repo (which becomes `Canasta-Go` after rename, archived).

Files added/changed:
- `bridge/main.go` (new) — the minimal program (one `func main()` with `fmt.Println`).
- `Makefile` — new `bridge` target cross-compiles to `dist/bridge/canasta-<os>-<arch>` for `linux/amd64`, `linux/arm64`, `darwin/amd64`, `darwin/arm64`. Uses `-trimpath -ldflags="-s -w"`.
- `.gitignore` — adds `dist/` (build output, not committed).
- `.github/workflows/release.yml` — build step now compiles the bridge from `./bridge` instead of invoking `build.sh`. Output path and asset names unchanged, so the existing upload-artifact and publish steps work without further modification.
- `VERSION` — bumped to `3.7.0`. On merge, this triggers `release.yml` which publishes the bridge binaries as release assets at `github.com/CanastaWiki/Canasta-CLI/releases/v3.7.0`.

Closes #739.

## Test plan
- [x] `go vet ./bridge` — OK
- [x] `go test ./...` — all existing tests pass (no regression)
- [x] `make bridge` — 4 binaries produced, ~1.5–1.6 MB each (~6.2 MB total)
- [x] Run `dist/bridge/canasta-darwin-arm64` on macOS — message prints correctly
- [x] Run with arbitrary args (`canasta upgrade`, `canasta version --help foo bar`) — same message regardless
- [x] Mirror the exact workflow command locally (`GOOS=linux GOARCH=amd64 go build -trimpath -ldflags="-s -w" -o build/canasta-linux-amd64 ./bridge`) — produces 1.5 MB binary at the expected path
- [ ] After merge: confirm the release workflow runs, builds 4 bridge binaries, and attaches them to a v3.7.0 GitHub release
- [ ] After merge: on a clean VM with the previous (3.6.3) Canasta-CLI installed, run `canasta upgrade`. selfupdate should follow the v3.7.0 download URL, install the bridge, and the next `canasta` invocation should print the migration message


cicalese:
Reconsidered the bridge-binary approach. The issue: when a 3.x user runs `canasta upgrade`, the bridge replaces their working `canasta` binary with a stub that only prints a migration message and exits. Every subsequent command (`canasta start`, `canasta backup`, `canasta list`) becomes inert. Their wiki ops are bricked until they manually run `curl … | bash`.

Switching approach: instead of a bridge stub, publish the **real** v3.7.0 Go CLI (functional, with a deprecation warning printed to stderr on every command) and use a cross-repo release mirror so legacy clients hitting `api.github.com/repos/CanastaWiki/Canasta-CLI/releases/latest` find a working binary after the rename.

New PR: #741 (`deprecation-warning-and-cross-repo-mirror`). Highlights:

- Adds `printDeprecationWarning()` in `cmd/root/root.go` — fires from `PersistentPreRun`, prints to stderr, suppressible with `CANASTA_SUPPRESS_DEPRECATION_WARNING=1`.
- Adds a cross-repo mirror step to `.github/workflows/release.yml` using a CanastaWiki GitHub App token (skipped when the App isn't configured or when the workflow is running from `CanastaWiki/Canasta-CLI` itself, to defend against post-rename self-mirror).
- No `bridge/` directory, no Makefile bridge target, no full-CLI build replacement. Build pipeline keeps producing the actual Go CLI.
- VERSION stays at 3.6.3 — bump to 3.7.0 happens in a small follow-up commit *after* the rename + App setup, so the release pipeline lands artifacts in the right repos (Canasta-Go regular publish + Canasta-CLI mirror).

Closing this PR without merging.

cicalese:
Closing without merging — superseded by #741 (cross-repo release mirror) per the pivot discussion in the prior comment. Branch left unmerged for reference.

